PR #29, opened 2025-03-10
Fixes Dracula theme BBCode colors

FatCatClient:
Though I mostly did this because the blue text was pretty illegible in this theme, these were always the intended colors instead of using the default ones. I just didn't get to push this fix upstream.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,17 +19,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fixes #15, ALinux Arm64 build are actually x86_64 [[Commit]](https://github.com/Fchat-Horizon/Horizon/commit/db43135677cb8e5e57f51bb0ffb417834ccd4103)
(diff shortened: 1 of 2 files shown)


CodingWithAnxiety:
Seems good. I'll do a more thorough review and merge here soon.

FireUnderTheMountain:
I'm trying to follow along but some of the @import chains seem a little weird. @FatCatClient Do you know what the `$blue-color: $blue-color;` in variables/dracula_derived is doing? An artifact of another era? I can't see how it actually has an effect on anything, but the other _derived variables also declare a blue
color, so I'm not sure what's up.

FatCatClient:
> I'm trying to follow along but some of the `@import` chains seem a little weird. @FatCatClient Do you know what the `$blue-color: $blue-color;` in variables/dracula_derived is doing? An artifact of another era? I can't see how it actually has an effect on anything, but the other _derived variables also declare a blue color, so I'm not sure what's up.

Since it's been a while since I wrote that, I can't say for sure. I assume I was just trying to match it for consistency's sake with the existing themes. Looking at the file history for those, they've been unchanged since the original development of 3.0 so I don't even know if declaring a random blue variable there is some load-bearing hack or a mistake that was just propagated since.

PR #42, opened 2025-03-11
Adds additional instructions for installation on MacOS

FatCatClient:
Tim Apple doesn't want people running unknown apps without using the terminal, so I hope that these instructions are clear enough for whoever else uses MacOS and doesn't know how to disable Gatekeeper (whether globally or on a per-app basis).

If we'd want a developer license, I think it costs $99 for 5 years. But it _would_ allow for released versions to be signed and therefore automatically approved.


PurpleDonuts:
Yeah. Enrolment into the Apple Developer Program is $99 US per Membership Year. [Source](https://developer.apple.com/programs/enroll/)

CodingWithAnxiety:
Made some changes. Verify and I'll merge.

FatCatClient:
If you bring back step 2, I’ll sign off on the changes.
I’m trying to keep the instructions as dummy proof as possible, and you need to have the app installed before you can tell MacOS to disable the certificate check.

CodingWithAnxiety:
Oh, did I remove step two? I might have without noticing in the merge editor. Give me a moment and I'll check it out

FatCatClient:
Since the app file itself is now also called ``F-Chat Horizon.app``, the CLI instructions need to be changed to reflect this different file name.

PR #43, opened 2025-03-11
Adds automatic update check

FatCatClient:
In all honesty, I can't be bothered to rewrite the whole spiel I did back on the old Rising repo on why the automatic updater wouldn't really work with the existing GitHub releases pipeline, as it would require a bespoke hosting solution for the downloads _(blah blah blah, multiple instances of the RELEASES file which Squirrel needs for patching and file name collusion blah blah blah)_

If this repo will move to a different version control system and use its own CI/CD pipelines, this intermediate solution might not be necessary but I'll propose it anyway.

Originally made for the old Rising repo, like most of the other PRs I've done over the past few days.


FireUnderTheMountain:
Window.vue 607

CodingWithAnxiety:
This looks to be a pretty quality commit, I'll have the fix merged later today.

CodingWithAnxiety:
This looks to be a pretty quality commit, I'll have the fix merged later today.

FatCatClient:
Horizon actually uses pre-release tags (unlike) Rising, so I'll retool this to use the commented out beta channel setting for those. Right now, merging this PR would mean people would be prompted to update to pre-release versions. Not a good idea!

PR #51, opened 2025-03-18
fix: Dracula theme background restored to intended color

FatCatClient:
The original theme required a quick hack to get the intended background colour, but with all the SCSS dependencies up to date and refactored, it over-corrected itself too brightly.


FatCatClient:
I think this is everything, no clue why this theme specifically didn't like the SCSS function refactor.

PR #53, opened 2025-03-18
Updates .gitignore for new release scripts

FatCatClient:


freenutsxd:
Would the same line need to be added to .prettierignore?

FatCatClient:
Unless any of the build scripts output file types which Prettier parses— like JS, TS, or markdown files, that won’t be needed.

FatCatClient:
image

PR #55, opened 2025-03-18
Fix zoom in hotkey + re-allow zooming out

### LINKED ISSUE #52: Zoom not working (anymore?)
Hello

Not sure since when it is like this, because the last time I used zoom was still with the Rising version, but zoom in/out or restting isn't working anymore. Tried to zoom out a bit, because the zoom was too big after the update 1.30.1 x64, and it unfortunately didn't work. OS is Windows 10 64bit.

freenutsxd:
Some Electron update made the existing Ctrl + Plus hotkey not work and made zooming out no longer possible, so these two quick fixes should resolve those issues.

Additionally, it looks like the overall font/zoom has been slightly increased overall, but I wasn't able to find a reliable way to globally fix that. Might be worth looking into if someone else wants to use this PR as a baseline.

Closes #52.

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -546,7 +546,7 @@ function onReady(): void {
         click: (_m: electron.MenuItem, w: electron.BrowserWindow) => {
           // log.info('MENU ZOOM-');
           zoomLevel = Math.max(
-            0,
+            -3,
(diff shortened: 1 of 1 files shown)


freenutsxd:
Closes #52.

PR #61, opened 2025-03-22
fix: lack of shadows in black (and other) user names

### LINKED ISSUE #45: Black custom name colours are pretty much illegible
The BBCode equivalent at least has a shadow for a border effect (although it's not particularly legible at smaller sizes either way). Meanwhile black user names just fade into the background... except on the full dark theme where you get this:

Image

Unless someone has a better alternative, my suggestion is that we just change the colour for both the BBCode tag and the custom username to the same one that all genderless characters use (#737373), since that one is actually legible.

FatCatClient:
I personally think it'd be cleaner if instead of this we just set the CSS rule for custom usernames and BBCode text to be one and the same. But since that robs people of the opportunity to potentially update a theme to use different colours (and this being the solution that was proposed by @CodingWithAnxiety  in #45 ).

Fixes #45.

### DIFF
--- a/scss/themes/variables/_dark_derived.scss
+++ b/scss/themes/variables/_dark_derived.scss
@@ -1,5 +1,6 @@
 $blue-color: #06f;
 
-.blackText {
+.blackText,
+.blackNameText {
   @include text-outline($gray-600);
 }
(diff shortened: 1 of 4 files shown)


CodingWithAnxiety:
Hi, 

It's hard for me to merge PRs since I don't have internet at the moment.

I'll be in contact with you, Cat, but I'll likely invite you into horizon sometime today and you can merge it yourself. 

I have faith in your abilities to make good patches.

PR #66, opened 2025-03-26
Fixes ChatView component not being mounted correctly

### LINKED ISSUE #56: 1.30.1 - Inability to re-order channels
Was unable to move characters or rooms in the side list, initially thought it was related to having a second tab open but closing it did not fix nor did restarting the client. Was unable to get any relevant log info for this issue.

### LINKED ISSUE #60: Saved ads in the ad editor don't load anymore
I'll try and narrow this down some other time, but the bug can be easily replicated:

1. Save an ad in the ad editor
2. Disconnect or close the tab
3. Log in again
4. Ad is now gone

The file _does_ save correctly, so I'm 99% sure it's just a problem with loading the correct setting upon connecting.

### LINKED ISSUE #65: Notification issue
Not talking about the Windows notification, but about the blinking icon when the Client is minimized to the tab bar.

It won't blink anymore, when the chat/room tab (within the client) is active/selected, and you were mentioned in there. So, let's say you are writing with "Random Username", and keep the tab active/selected, so you can read their messages. Then you minimize the whole Client (or maxmize your browser to look something up for example - while the chat with "Random Username" is still selected). When "Random Username" writes a message, the icon won't blink.

It will only blink, when you were texted or mentioned in a non-selected or non-active chat/room tab.

### LINKED ISSUE #67: Alt Commands to move between channels don't work
When using Alt+up/down on the base 3.0 client, you can move between channels and DMs (like in Discord), but for some reason this doesn't work in Horizon. Is this intended?

### LINKED ISSUE #71: Feature Request - Restoration of navigation hotkeys from Rising
In F-chat Rising, I could use alt+Up/Down arrow to scroll between chat tabs (i.e.: private rooms, DMs). Is this something that we would be able to put in Horizon?

FatCatClient:
Fixes #65 
Fixes #60 
Fixes #56
Fixes #67 (and fixes #71)
Also fixes an unreported bug where the idle timer doesn't work (and basically everything the chat component is supposed to do on load), and font size settings not being initialized on page load.

### DIFF
--- a/chat/ChatView.vue
+++ b/chat/ChatView.vue
@@ -274,10 +274,9 @@
 </template>
 
 <script lang="ts">
-  import { Component } from '@f-list/vue-ts';
-
   import Sortable from 'sortablejs';
 
+  import { Component, Hook } from '@f-list/vue-ts';
   import Vue from 'vue';
   import { Keys } from '../keys';
   import ChannelList from './ChannelList.vue';
@@ -351,7 +350,8 @@
     privateCanGlow = !this.channelConversations?.length;
     channelCanGlow = !this.privateConversations?.length;
 
-    mounted(): void {
+    @Hook('mounted')
+    onMounted(): void {
       this.keydownListener = (e: KeyboardEvent) => this.onKeyDown(e);
       window.addEventListener('keydown', this.keydownListener);
       this.setFontSize(core.state.settings.fontSize);


astrayblackcat:
This breaks so many things it should probably be merged asap to release a hotfix.

Issue #37, opened 2025-03-10
Ads showing up twice in Recon/Ad History


freenutsxd:
I reported this issue in Rising as well, but randomly there will be times where ads show up twice within the character screen's Recon tab and when you check their ad history with the Ads button. When a single ad is posted by a character, that message will show up twice in the history. It seems to happen a majority of the time, but rarely it won't occur, not really sure why/how it happens.

![Image](https://github.com/user-attachments/assets/59f5c929-c32f-4cc8-9d5d-aa454d122743)

![Image](https://github.com/user-attachments/assets/70d3c6ed-19ea-4926-8e85-5651333eef40)

PR #72, opened 2025-04-01
fix: Duplicate event listeners

astrayblackcat:
Cache startup was being run more than once, it seems more correct to have that function called when a connection is established ([here](https://github.com/Fchat-Horizon/Horizon/blob/f798c39b57d067d55646e54c78cbb626bc33a9c4/chat/Chat.vue#L277)) rather than when this is called, otherwise relogging will break them.

Also fixes the handling of duplicate event listeners in the event bus, although this should not strictly be necessary and the solution in place is rather fragile. Comparing two functions like this will never have equality unless converted to a primitive. This probably wants a refactor, realistically...

Closes https://github.com/Fchat-Horizon/Horizon/issues/37.

P.S. First contribution, woo.

### DIFF
--- a/chat/preview/event-bus.ts
+++ b/chat/preview/event-bus.ts
@@ -69,7 +69,7 @@ class EventBusManager {
 
     this.eventCallbacks[eventName] = _.filter(
       this.eventCallbacks[eventName],
-      cb => cb !== callback
+      cb => cb.toString() !== callback.toString()
     );
   }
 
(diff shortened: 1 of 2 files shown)


FatCatClient:
Seems good to me, but I'll hold off on merging it myself until the current batch that has been merged into the dev branch has been approved for merging into the beta branch.

FireUnderTheMountain:
Just curious, have you deciphered the meaning of the `// intentionally background` comment in Index.vue right below the removed init? It's always bothered me, making me think there's an actual design to that section that I don't fully grasp.

astrayblackcat:
Pretty sure it's an async function not being awaited and the comment is saying that it not being `await` is intentional.

PR #75, opened 2025-04-03
fix: Remove unnecessary include line from tsconfig

astrayblackcat:
The default include grabs all valid files in any directory nested to any degree, so this include is actually *excluding* all other files, which seems to be fine at build time(?) but it makes LSPs in editors very upset with a ton of errors that would be handled by tsconfig.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -13,6 +13,5 @@
     "allowSyntheticDefaultImports": true,
     "esModuleInterop": true,
     "experimentalDecorators": true
-  },
-  "include": ["./electron/main.ts"]
+  }
 }


astrayblackcat:
Thinking about it, I think this is from the times where you had to compile to js before building with Electron, so no wonder it doesn't break anything.

FatCatClient:
Wait-- which LSP errors does this solve again? Maybe I'm lucky (or am still using an outdated version of the TS LSP :^) ) but mine doesn't seem to be giving me a ton of errors while working on this repo.

astrayblackcat:
Huh, that's odd.  I'll have to check later to see what exactly it was, but off the top of my head I was getting errors with decorators because the LSP wasn't reading the line enabling them.

I also tried a couple versions; the latest and the version of TS used by the repo, and they both threw a varying amount of errors for me, although the latter just has trouble with Vue for me.

I'm pretty sure I'm just using the vscode TS LSP, but I am running it in neovim. It works completely fine in other situations but it's not impossible it's on my end.

FatCatClient:
It could be on your end; I'm using [vls 0.8.5](https://www.npmjs.com/package/vls/v/0.8.5) and [typescript-language-server 4.3.4](https://www.npmjs.com/package/typescript-language-server/v/4.3.4) both as globally installed packages with NPM and with Neovim using Mason LSP.

astrayblackcat:
Yeah, those are quite a bit older than what I'm running, which might be more appropriate considering how out of date the previous maintainer of the repo let these packages get, lol. 

It seems like it's a problem with the vscode TS LSP (vtsls) specifically. It seems to depend more on tsconfig than other LSPs. The problem also persists in actual vscode, which isn't surprising because it's the same LSP.

Which implies anyone who wants to contribute and codes in vscode would have the same issue unless they do some manual config. And to that end, we're not actually compiling with `tsc`, so I'm *fairly* sure there's any real benefit in keeping the line around anyway (although I may be wrong)

astrayblackcat:
I actually haven't seen that, though I suspect deleting the whole tsconfig might break some other things with LSPs because they depend on certain flags to determine things like import syntax and what functions are permitted based on the compile target. 

From memory, VSCode sometimes likes to throw an error about large folders not being excluded, so maybe that's the problem. We might need to actually be more specific with our include in that case... (but less specific than it was originally, lol)

astrayblackcat:
I do get the error without it set, but setting the include to `["./*"] seems to work. Might need more looking into than I thought.

FatCatClient:
I'm not exactly a Typescript expert, so this might be a bit of a shot in the dark- but if the tsconfig file takes a string array, can't we just try and set it to include the folders that we actually use? Or do we _need_ the node_modules folder included?

astrayblackcat:
Should just be able to exclude node_modules, that's probably the issue. It's not necessary to be included, to my knowledge.

astrayblackcat:
Does `"include": ["*/*"]` with `"allowJs": false` work for you? That works perfectly on my end. Should include everything necessary... I have no idea why that works but the default doesn't, but I'm not going to question it.

FatCatClient:
With the following settings in tsconfig.json it seems to work fine(?). It even stops adding the function definition to the list of references, which it kept doing earlier. 

```json
{
  "compilerOptions": {
    "target": "es2017",
    "module": "commonjs",
    "sourceMap": true,
    "allowJs": false,
    "noEmitHelpers": true,
    "importHelpers": true,
    "forceConsistentCasingInFileNames": true,
    "strict": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "allowSyntheticDefaultImports": true,
    "esModuleInterop": true,
    "experimentalDecorators": true
  },
  "exclude": [
    "./node_modules/**",
    "./electron/node_modules/**"
  ],
}
```

_Maybe_ we could also exclude ``./electron/app/**``, but I tested this right before I saw your edits.

astrayblackcat:
Oh, yeah, it's probably ./electron/node_modules/ doing it, actually... good catch.

PR #86, opened 2025-04-10
feature: Dot for original color when using custom color

BootsieWootsie:
Puts a dot beside people with custom colors for their names so you can still easily spot the gender while preserving their custom name color. Appears in chat, sidebar, and at the top of private conversations.

Has included option to enable/disable it. Currently set to default to on.

Had to expand the sidebar by 10 pixels to accomodate a name with it that was very long.

### DIFF
--- a/chat/SettingsView.vue
+++ b/chat/SettingsView.vue
@@ -696,8 +696,7 @@
         settings.risingShowHighQualityPortraits;
       this.horizonShowCustomCharacterColors =
         settings.horizonShowCustomCharacterColors;
-      this.horizonCCCOriginalDot =
-        settings.horizonCCCOriginalDot;
+      this.horizonCCCOriginalDot = settings.horizonCCCOriginalDot;
 
       this.risingFilter = settings.risingFilter;
 
--- a/chat/SettingsView.vue
+++ b/chat/SettingsView.vue
@@ -334,6 +334,17 @@
       </div>
 
       <div class="form-group">
+        <label class="control-label" for="horizonCCCOriginalDot">
+          <input
+            type="checkbox"
+            id="horizonCCCOriginalDot"
+            v-model="horizonCCCOriginalDot"
+          />
+          Show a dot with the original gender colour
+        </label>
+      </div>
+
+      <div class="form-group">
         <label class="control-label" for="risingCharacterTheme">
           Override UI theme (for this character only)
           <select
@@ -631,6 +642,7 @@
     risingShowPortraitInMessage!: boolean;
     risingShowHighQualityPortraits!: boolean;
     horizonShowCustomCharacterColors!: boolean;
+    horizonCCCOriginalDot!: boolean;
 
     risingFilter!: SmartFilterSettings = {} as any;
 
@@ -684,6 +696,8 @@
         settings.risingShowHighQualityPortraits;
       this.horizonShowCustomCharacterColors =
         settings.horizonShowCustomCharacterColors;
+      this.horizonCCCOriginalDot =
+        settings.horizonCCCOriginalDot;
 
       this.risingFilter = settings.risingFilter;
 
@@ -787,6 +801,7 @@
         risingShowPortraitInMessage: this.risingShowPortraitInMessage,
         risingShowHighQualityPortraits: this.risingShowHighQualityPortraits,
         horizonShowCustomCharacterColors: this.horizonShowCustomCharacterColors,
+        horizonCCCOriginalDot: this.horizonCCCOriginalDot,
 
         risingColorblindMode: this.risingColorblindMode,
         risingFilter: {
--- a/chat/UserView.vue
+++ b/chat/UserView.vue
@@ -12,6 +12,8 @@
     @click.right.passive="dismiss(true)"
     @click.left.passive="dismiss(true)"
     ><img v-if="!!avatar" :src="avatarUrl" class="user-avatar" /><span
+    v-if="!!cccDot" :class="cccDot"></span
+    ><span
       v-if="!!statusClass"
       :class="statusClass"
     ></span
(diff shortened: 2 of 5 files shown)


FatCatClient:
This sounds cool and useful-- could you add a screenshot real quick so I don't have to build it locally to see what it looks like?

BootsieWootsie:
Of course, here ya go!

For PMs

For Rooms


Also works on the sidebar. It looks the same there as it does on those two, though. It should work on searches as well too, no reason it shouldn't to my knowledge since they should be inheriting the same way the original custom colors did and they all inherit from UserView.vue to some extent.

The way things are integrated right now, I'm not wholly sure it'd be possible to remove it from any single one though or make it optional on a case-by-case basis without a good wee bit more work.

CodingWithAnxiety:
This needs a bit of work, but I think this change is a step in the right direction. I'll probably make a patch for this pull request once I get home from work, leaving open for the time being.

astrayblackcat:
I like the idea of using gender markers — I don't remember if the client has any sort of colourblind support right now, so that could be a nice improvement outside of just custom name colours.

astrayblackcat:
Oh, so we do. It does just change colours, though... I think it would still be helpful if there were an at-a-glance way to get that information that didn't rely on colour at all.

C-Kawinski:
Think it only changes the colors within the theme currently but i could be mistaken. @CodingWithAnxiety  would know

BootsieWootsie:
> Think it only changes the colors within the theme currently but i could be mistaken. @CodingWithAnxiety would know

As far as I can tell, yeah. There's a theme file for the colourblind colours that just overwrites the base colours for genders on all themes. Potentially means some incompat with other themes that'd make certain colours in colourblind hard to see.

> I like the idea of using gender markers — I don't remember if the client has any sort of colourblind support right now, so that could be a nice improvement outside of just custom name colours.
> 
> edit: The problem of course becomes picking out markers...

The marker choice is decently limited at least unless a new font is being added. Barely enough with a little overflow for a few alternate choices. Browsing through, don't actually have enough to leave each gender unique unless using the double symbols, which would put us just one ahead of the list or leaving people without a marked gender without a symbol or giving them something rather out there.

FatCatClient:
The FontAwesome set has a couple of additional gender icons (at least, these are the results I get when I search for the word "gender"). 
image

But in all honesty I don't know if something like the stroked Mars and Mercury signs already have a distinct meaning in relation to gender. It's a sensitive topic and, in my personal opinion, I think it's better to just have a circle matching the regular colours than to accidentally offend someone.

Or maybe I'm just being overly cautious 🤷

BootsieWootsie:
Yeah, pretty much no getting around it with the gender options available on the site. Almost certainly going to have some level of arbitrariness too. Especially considering that in my current little test I'm having to swap a few for visual distinction and ignore any kind of existing convention largely.

Just gonna plonk a few images down, it's somewhat attempted to have a reasonable picking but there's a lot of arbitrariness and I did specifically choose some things cause they looked visually distinct enough.
eee fff ggg

(Wasn't sure about posting the names of random peeps on here, so overzealous cutting)

FatCatClient:
I can't speak for the specific meanings nor the implications they'd have for the gender options on the site, but I do think this looks very clean and visually distinct. It makes for a good accessibility option too, for people who might have trouble telling the colours apart (or are simply colour blind).

astrayblackcat:
It does look good, I won't argue with that. But it might be a bit too opinionated. The gender options on the site are basically just a weird mashup of gender identity and the character's sex. And there might be more than one interpretation of them in quite a few cases which mean it might tread on some toes somewhat.

BootsieWootsie:
Could choose more arbitrary symbols if that's a big enough concern, things like the dice and similar squared symbols but might run into issues with it's size and visibility as a consequence. 

Biggest reason I didn't use the mars symbol with the stroke in the normal mars direction was just cause it would be harder to see.
Letters would be simple, with the exception of M since it's used twice, would also have to put some form of barrier like a small dot between it and the actual name. Basic shapes are an option too. 

Big issue with anything other than letters is that things would have harder recognition if they look too similar since the icons can be rather small and obscure finer details and there isn't an instant basis to recognise that meaning in your own head, would have to learn it.

SlimCat achieved it largely with two-tones on it's gender icons, we could pick that idea up since there's nothing really stopping us doing that and it would allow us to effectively cut down to 4 icons. A mars, venus, something for none, and a trans symbol. Obviously though, on the accessibility front it could end up still causing issues doing it that way.

Also, should I push the version of this I have on my end to this PR or make it a new one?

astrayblackcat:
The problem with letters is that it'd need to be a monospace font, or else you end up with weird vertical alignment issues. But it would probably be the safest, solely as an optional colourblind mode option, whilst you go with some other icon outside of colourblind mode.

BootsieWootsie:
> The problem with letters is that it'd need to be a monospace font, or else you end up with weird vertical alignment issues. But it would probably be the safest, solely as an optional colourblind mode option, whilst you go with some other icon outside of colourblind mode.

Way things are set up in my current version, using fontawesome's letters should be alright since it just makes the span for the icon a fixed width so that where names start and end are consistent across the board. Leaves a tiny amount of dead space with some symbols, shown above, but it felt odd having the names start at different spots depending on gender.

We don't currently have the SVGs of those letters available to us since we're on 5.15.4, but could just update it to 6.7.2 if we want them in the SVG form and letters is where things settle.

BootsieWootsie:
Put a new PR on with the gender marker idea since it seemed to be more liked than the dot was and it does look a good deal better with the added bonus of potential colourblind assistance.

BootsieWootsie:
Just gonna quickly close this while I remember since the markers is far better done than this and looks better.

Issue #73, opened 2025-04-01
Large log files reduce performance(?)


FatCatClient:
I haven't really diagnosed this properly, but I've found that on characters where certain room logs are pretty big (ie over 20 MB), I'll feel very noticable stutters and input lag all around (over 400ms if the dev tools profiler is to be believed). Leaving these rooms seems stop this from happening.

Digging a bit in the code, in [electron/filesystem.ts](https://github.com/Fchat-Horizon/Horizon/blob/f798c39b57d067d55646e54c78cbb626bc33a9c4/electron/filesystem.ts#L342) it seems that it _does_ append the file, which isn't what I thought it'd do-- so maybe some further research is needed to profile whichever function is causing it to hang.
We at least can see that it doesn't write the files asynchronously, so it's not implausible that writing the files itself is what's causing these hiccups.

FatCatClient:
Marking this as a dupe for the other report because that one has actual discussion going on.

PR #87, opened 2025-04-11
Asynchronous log handling

FatCatClient:
- [ ] Handling files
    - [x] Writing
    - [ ] Reading
- [ ] Testing

See #73, I pooped this out really quickly during my lunch break so I'll need to spend some more time on this.

### DIFF
--- a/electron/filesystem.ts
+++ b/electron/filesystem.ts
@@ -22,15 +22,15 @@ declare module '../chat/interfaces' {
 const dayMs = 86400000;
 const read = promisify(fs.read);
 
-function writeFile(
-  p: fs.PathLike | number,
+async function writeFile(
+  p: fs.PathLike,
   data: string | NodeJS.ArrayBufferView,
   options?: fs.WriteFileOptions
-): void {
+): Promise<void> {
   try {
-    fs.writeFileSync(p, data, options);
+    await fs.promises.writeFile(p, data, options);
   } catch (e) {
-    remote.dialog.showErrorBox(l('fs.error'), (<Error>e).message);
+    remote.dialog.showErrorBox(l('fs.error'), (e as Error).message);
   }
 }
 


CodingWithAnxiety:
Keep me in the know. I've benchmarked and noticed no issues, but I'm also on Linux.

FireUnderTheMountain:
I'm curious, of course.. did this fix the issue? And just as importantly, what sort of disk are the logs on, solid state or spinning rust?

FatCatClient:
I haven't tested it on my Windows machine yet, but I doubt it's some hardware-related issue because my PC has a Samsung 980 NVME drive and a Ryzen 5 7600, which I think should be more than enough. The computer where these hitches don't happen is a Macbook Air M2, which in every measurable way doesn't match my desktop's performance.

It's also not a "consistent" low responsiveness, if that makes sense. There are just noticeable hitches in responsiveness for characters with lots of "big" logs (between 16 and 25 mb), which seem to disappear if I leave those rooms or delete the logs. These hitches show up in the dev tools performance tab under the Interaction To Next Paint header, where they're usually around 200 to 400 ms.

I'll build a Windows version for this branch and restore the bigger logs so I can do the same test there, but if I get similar results for performance while measuring the main branch's version of the ``logMessage()`` function I'd be stumped.

FireUnderTheMountain:
I remain curious, because if append is already the write method, then (like you pondered earlier), there shouldn't be a lot of writes that punish large files more than the others. If it **is** blocking on writes, that would be caused by being in too many channels/doing too much logging, no? Do you log ads by any chance? I definitely don't.

FatCatClient:
I think I'm going to close the PR for now until I have a better idea of what the problem might be. Once I get the time to try and narrow down the problem (and see if this even happens for other people), I'll have a solution that I can actually feel confident in.

PR #88, opened 2025-04-12
feature: Gender markers beside names

BootsieWootsie:
Putting this to it's own PR since this is functional and working and looks a good bit better than the colour dot I went for earlier in https://github.com/Fchat-Horizon/Horizon/pull/86 and has a slight bit more functionality.
It is also more tested, been using it without issue for a good portion of the day. 
I still need to run Prettier on the changes to UserView but I believe it is in line with the requirements.

Current icons used are primarily mars, venus, the trans symbol, and arbitrary versions of those that I felt were visually distinct enough to make things visible while still somewhat matching the base gender associated where possible. 
It defaults to disabled but has proper implementation in the Horizon settings menu and is easy to toggle at will. The setting change is not currently watched so you have to change channel for the changes to take effect.
These symbols appear in chat (rooms, PMs, console), the sidebar, recent conversations, and search.
This version does not change the colours of these icons on custom names, the icon itself is enough indication.

Preview of currently used symbols
eeefffggg
Sorry for the terrible crop, didn't think it was ideal to include random profile names over on github.

There was discussion of changing them to things that are potentially less of a contestable or contraversial choice.
Current suggestions I can recall are 
- Basic Shapes
- Letters
- Less symbols but two-tone them like SlimCat's version of this feature

Technically, all of these options including the current setup could be done. The two-tone version may be harder to make toggleable but everything else could be done with a simple dropdown and just let people choose which they want.
Letter suggestion has two genders that overlap, other than that it could work alright.

This currently also includes an offline symbol for when someone is offline because anyone who has not logged in yet gets marked as genderless. This symbol is currently a times symbol and it changes the moment someone goes online or offline due to how the UserView is set up outside of this change.
The colour of a name does not change when this occurs so they retain their original gender colour.

Unlike the previous PR, this does not include any changes to _chat.scss


astrayblackcat:
I forgot to mention in the other PR, but here's probably a better place to mention it: I'd probably suggest swapping the two purples around, since the Mercury symbol is typically used to mean intersex, so it lines up perfectly. And I think the venus-mars one would probably be more obviously understood in that case.

BootsieWootsie:
True, I was somewhat hesitant to do it for consistency with dark blue but... Yeah, it probably would be a little more appropriate thematically to swap them.

astrayblackcat:
I was a bit against it before, but I'm coming around on it as an accessibility option, mostly. There's no real way around the existing issues with genders on the site, since it's kind of a top-down issue we don't have a hand in. And the accessibility is a nice enough benefit that I think any potential issues can be overlooked, since they're issues that ultimately lie with the actual site.

BootsieWootsie:
Yeah, only really so much that can be done to mitigate that front sadly. Made that little change there!

BootsieWootsie:
Oh, seems there was a minor issue I didn't notice. You should relog the first time after you turn the feature on or some of your friends and bookmarks list may not gain the marker. Simplest fix is probably just putting that info into the text for the setting.

FatCatClient:
I want to say that it's probably preferable if it works normally, but it turns out that the same thing happens with custom character colours too (which has the same issue in conversation view).
Toggling seems to work fine with avatars, but since those don't show up in the user list it can't break there. Could you check the avatar setting for the UserView component and figure out why it works there (and hopefully get it to work the same for your setting)?

FatCatClient:
Right, that's around what I expected-- except that I assumed it was a property on the UserView component. 

I'm not sure whether we should consider fixing the names not toggling to be out of scope for this particular PR, but that should probably just be a [v-bind](https://vuejs.org/guide/essentials/class-and-style) either way.

BootsieWootsie:
I was looking in the wrong place after that cause I was looking at a wholly different setting than the one I thought. risingShowPortraitInMessage was the one I was meant to, it is used in message_view.ts where UserView's element is created using it as a boolean for a prop.

It toggles in chat nice and easy now, just added a prop for it and did it the exact same way as the image.

Sidebar is being more tricky, status icon can change there but that is done wholly inside UserView. Doing it with a watch and making it update when the setting changes doesn't appear to work though.

BootsieWootsie:
So, I found a solution to this using get, I need to give it a little more testing but it toggles properly on every possible component that accepts it on my test version.

BootsieWootsie:
Replaced the get component on each individual UserView that the previous change made with ones put into slightly better places. UserList now grabs it and passes the data down to each UserView instead, it's grabbed the same way as avatar in message_view, bbcode grabs it itself now too.

Downside is that it doesn't update on bbcode [user] components instantly in the channel that you are looking at and that searches need to be redone after changing the setting too.

FatCatClient:
I haven't had the chance to look too deep into the implementation of this, nor merge and test it locally. But is the X symbol for offline names a placeholder for characters whose gender hasn't been cached yet, or does it shift back into an X if somebody logs off?

BootsieWootsie:
Yeah, it will shift into an X and back again whenever someone goes offline and online. If I recall, my implementation before the X it would shift them to the icon for being genderless whenever they went offline so I just included a quick check for status too. 

Looking back on it now, calling for character.overrides.gender instead of character.gender would likely change that to use the cached version rather than setting them to genderless the moment they go offline. I admittedly have rather grown to like having the offline symbol though (as well as utilising my option to make names go back to grey when people go offline) as it gives you a much quicker indicator inside of rooms when people are offline.

I could fix it up to potentially do both, calling for their override gender when not using the offline name colour change and using the old value if you are using that option.

FatCatClient:
I think we should just go with the implementation as is, and if people would rather have it work the same way non-custom name colours so we can always change it.

BootsieWootsie:
Fucked up a little while trying to revert an accidental change to the adjustable-autoreply. Should be fixed now, sorry about that!

PR #95, opened 2025-04-15
fix: Listeners are not properly removed when logging out

astrayblackcat:
Normally, trying to add an event listener to the DOM that already exists fails, but adding one in a Vue component mount hook creates another instance of the function, bypassing the built-in duplicate check.

When logging out, the event listeners added here aren't removed properly because logging out only unmounts it, it doesn't trigger the destroy hook. Vue 3 has an onUnmounted hook for this, but this is the best we have.

There might be other solutions to this, like removing any others before we add them or having some variable keep track of them, but this seems the simplest and least messy.

...event listener bug 2: electric boogaloo!

### DIFF
--- a/chat/ChatView.vue
+++ b/chat/ChatView.vue
@@ -464,6 +464,7 @@
       void core.adCenter.load();
     }
 
+    @Hook('destroyed')
     destroyed(): void {
       window.removeEventListener('keydown', this.keydownListener);
       window.removeEventListener('focus', this.focusListener);


astrayblackcat:
Should also note it may be worth keeping an eye out for this happening in other places, because some of them are subtle or don't actually cause any issues. But anywhere a listener is added in a mount hook should be double checked. This one in particular was just very obviously breaking some hotkeys after a relog, so I fixed it.

FatCatClient:
Do you think it's only this specific component that doesn't call the destroyed() hook function on logout, or is it a problem throughout the app? In [Editor.vue](https://github.com/Fchat-Horizon/Horizon/blob/f798c39b57d067d55646e54c78cbb626bc33a9c4/bbcode/Editor.vue#L246) for example, there's an event listener that appears to be removed appropriately in the destroy hook, but if this method isn't being called for any of the elements this method call is useless. And that's just me checking the first instance of ``window.removeEventListener()`` I found.

Based on your findings you'd think that it doesn't destroy them properly, but if I am to believe the [Vue 2 documentation](https://v2.vuejs.org/v2/guide/conditional.html) it should destroy its children on disconnect, since [Chat.vue](https://github.com/Fchat-Horizon/Horizon/blob/f798c39b57d067d55646e54c78cbb626bc33a9c4/chat/Chat.vue#L13) is using a V-if directive, which allegedly "ensures that event listeners and child components inside the conditional block are properly destroyed and re-created during toggles."

I think it's worth looking into migrating to Vue 3 anyway, since apparently Vue 2 has been EOL since December 2023. But that's obviously out of scope for this PR.

astrayblackcat:
I just figured it out, actually! I was wondering why the lifecycle hook was just being completely ignored, but it's the same reason why ChatView wasn't mounted properly. We have a library for lifecycle hooks that's not standard or used anywhere else, it was created just for this.

https://www.npmjs.com/package/@f-list/vue-ts

Essentially, any instance with a lifecycle hook needs to have the hook decorator added before it, or it just doesn't get called at all.

Migrating to Vue 3 would probably be the better long-term solution, here, though. I'll update this PR with the proper fix, lol.

FatCatClient:
> Migrating to Vue 3 would probably be the better long-term solution, here, though.

There's thankfully a [compatibility library](https://www.npmjs.com/package/@vue/compat) that should help smooth out the transition, but even then going through all the instances of deprecated usages would be enough of an undertaking for me to want to wait until this is merged to give it any serious thought.

FireUnderTheMountain:
I'd like to try to test that this change has a meaningful effect. Is there a method or scenario you already use that produces a memory leak?

astrayblackcat:
Technically it's a memory leak, but it's really just an orphaned event listener with no way for it to be removed. Very minor, but it causes a bug with the channel/conversation scroll hotkey

Simplest test for this is to use alt+up/down while logged in, observe that it scrolls normally. Now, relog without closing the tab (same character or another, just don't close the tab), and re-try. You can also look in the dev console in Elements > Event Listeners and see that it's duplicated.

With the hook added, it properly calls the `destroyed()` function.

FireUnderTheMountain:
I knew this behavior was fishy so I went digging. Found that an over-excited editor [removed this line](https://github.com/Fchat-Horizon/Horizon/commit/bb4a760adaccf7e9dfdf364301327f701b126a49#diff-5bc252a7755fce3b5711085b600373289ac2e1267a600426bf4d63e11f6f331cL462) previously; so yes, this fix is exactly what's in order, thanks for being willing to find it. I clearly don't use this app to its maximum potential...

astrayblackcat:
It's an easy mistake to make, I think. Without reading that very specific library's documentation there's no warnings that they're necessary and it fails silently.

FatCatClient:
@astrayblackcat Do you happen to have a user-facing bug or issue this would fix from the top of your head? Asking for the changelog.

astrayblackcat:
> Do you happen to have a user-facing bug or issue this would fix from the top of your head?

When you relog/DC, the alt up/down keys to scroll conversations start skipping over slots equal to the number of relogs.

PR #99, opened 2025-04-16
Work in progress: Vue 3 migraton

FatCatClient:
It'll be out when it's out, and we like that. 

I can't even tell if this is building correctly at the moment-- if I alias vue to vue/compat in the main config, it starts fine and that has me worried. If I add it to the render config, it gives me a completely empty screen. Which I'm more inclined to believe is what I'm supposed to see like this.

But we're only 472 days overdue for Vue 2's EOL and it's 10 PM on a workday as I'm doing this, so what's a few days or weeks more? 😬

### DIFF
--- a/electron/package.json
+++ b/electron/package.json
@@ -76,10 +76,16 @@
     "build:mac": "mkdir -p dist && (node ../webpack production && electron-builder --mac dmg --universal) | tee dist/build.log",
     "build:linux": "mkdir -p dist && (node ../webpack production && electron-builder --linux deb tar.gz AppImage --arm64 --x64) | tee dist/build.log"
   },
+  "dependencies": {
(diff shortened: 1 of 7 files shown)


BootsieWootsie:
It might be worth taking the opportunity during the process of this to rename old Rising vars to Horizon ones since when this is implemented, not much will go in along with it that would really touch those.

FatCatClient:
Ignore the current build state in Github actions-- this most definitely does _not_ build. But it does actually build as a Vue 3 app now, whereas the things I described in the PR description were actually just 2.6.1.

Components need to be updated to the Vue 3 syntax, but I have to try and figure out if that's also part of the Vue 3 Migration Build or if it's not being aliased correctly right now.

But this is _technically_ progress.

PR #106, opened 2025-04-21
fix: Channel message lag

### LINKED ISSUE #102: Very laggy v1.30.2 and at current state, not usable
Not sure if it is just me, but the Client became very laggy. Especially when minimizing it, and then maximizing again, it freezes a few moments before it works again (being black at first, as if it has to re-render the whole thing from start again). CPU usage reaches always up to 25% (and is around 15%+ at default). Tried it with and without hardware acceleration, both same results. Scrolling the opted-in room list ot chat window is also very laggy. But even when not minimizing, it doesn't take long until the lags take over and make the Clien unusable.

In v.1.30.1 the CPU usage is usually between 1% and 6%, depending on what  exactly you are doing. I assume one of the function fixes is causing these performance issues. Unfortunately my knowledge in this matter is not enough yet to pinpoint the exact cause. I hope my hints are useful enough.

OS: Win10 64bit 22H2, latest Updates

And thank you for the efforts.

astrayblackcat:
Closes #102.

This probably needs further testing but I don't have time now. But I can't see that the removal `deep: true` has any negative effect (or any effect at all) and the lag problem is gone with it removed.

### DIFF
--- a/chat/ChatView.vue
+++ b/chat/ChatView.vue
@@ -356,25 +356,17 @@
       window.addEventListener('keydown', this.keydownListener);
       this.setFontSize(core.state.settings.fontSize);
 
-      this.$watch(
-        'conversations.channelConversations',
-        newVal => {
-          if (newVal?.length) {
-            this.channelCanGlow = false;
-          }
-        },
-        { deep: true }
-      );
+      this.$watch('conversations.channelConversations', newVal => {
+        if (newVal?.length) {
+          this.channelCanGlow = false;
+        }
+      });
 
-      this.$watch(
-        'conversations.privateConversations',
-        newVal => {
-          if (newVal?.length) {
-            this.privateCanGlow = false;
-          }
-        },
-        { deep: true }
-      );
+      this.$watch('conversations.privateConversations', newVal => {
+        if (newVal?.length) {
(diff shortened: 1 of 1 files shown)


FatCatClient:
I’ll be checking the Windows and Mac builds either tonight or tomorrow.

astrayblackcat:
Sounds good. I only got to very briefly test it but I couldn't get any stutters to happen with this and everything worked normally from what I can see.

But it's probably better tested to see if it fixes it for @0lm

FatCatClient:
And it explains why I assumed it coincided with large log sizes: those were all logs from very active rooms (like the main Fantasy channel), which would resolve itself once I left the room.

astrayblackcat:
Great to hear. I knew something was weird when it worked fine in 1.29. Just had to check the commit history on the file and there it was.

Doesn't seem like it breaks anything either so I'll call this one good.

astrayblackcat:
Cleaned up the written-right-before-bed commit message, should be all good to merge now if you want to hotfix this.

CodingWithAnxiety:
Hindsight being 20/20,, I don't know why I added deep: true. Merging.

PR #114, opened 2025-04-23
1.30.3 Release

### LINKED ISSUE #31: Bug: Author mispelling
Author in package.json has   "author": "The F-List Team and FChat Horizon Cobtributors", instead of Contributors.

### LINKED ISSUE #64: Missing Linux Builds
Clicking on the links to download on the main page directs to a 404 error. The only exception is the tar ball download.

### LINKED ISSUE #102: Very laggy v1.30.2 and at current state, not usable

### LINKED ISSUE #105: MacOS dmg still shows version 1.30.1
It appears the update was successful as the bugs around text size and ads disappearing is resolved, but the version is still listed as 1.30.1

Image

FatCatClient:
Closes #102 
Closes #105 
Closes #64 
Closes #31 

Changelog and version increase all included, but I set the release date for the patch to the 25th of April, so now I'll look like a dork if we merge it at a different time without changing that :^)

Pre-release builds are [here](https://github.com/Fchat-Horizon/Horizon/actions/runs/14626706215#artifacts).


CodingWithAnxiety (approved):
Approved, especially the issue templates. Feel free to merge this for me, FatCatClient.

Thank you all for your help!

PR #117, opened 2025-04-25
Fix scary "Logs corrupt! Repair now!" 

FireUnderTheMountain:
There is a problem in the [PM Partner Adder](https://github.com/Fchat-Horizon/Horizon/blob/main/chat/PmPartnerAdder.vue#L42) where if you enter a blank value and hit the button, you get a scary "logs are corrupt! repair now!" notice as a result of opening a PM with an empty-named character. The notice appears to be the result of trying to open a log file with no name.

There does not appear to be _actual_ corruption from simply opening the chat pane; but I haven't tried sending a message to see if bad things happen when trying to write to the non-existent log; probably not, since the user with no name cannot be online and thus you can't send them any messages. But a normal user shouldn't be ending up in this position in the first place.

This PR is the minimum necessary code to fix the jumpscare. There are only two places I've found that allow raw user-input for a character name, but there may be more. The `/priv` chat command is the yet-mentioned. I opt to fix the issue directly at the sources instead of in `core.characters.get()` function on three beliefs:
1. Changing `.get()` to be able to return null and hunting down all the places that might affect requires an amount of work greater than none.
2. Other potential changes (allow `.get()` to turn an empty string into a predetermined name, for example) require additional consideration that would delay a fix.
3. Sanitizing external input as soon as it enters the application is best practice.

There are additional problems with the surrounding code in both places (`if Character is not Falsey ...`? Really? The entire PmPartnerAdder is completely broken...) but they don't turn an innocent roleplayer's heart to ice by invoking the terrifying "logs corrupt" message, so they don't need to be addressed here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,11 +21,11 @@
 
 # Download
 
-[Windows x64](https://github.com/Fchat-Horizon/Horizon/releases/latest/download/F-Chat%20Horizon-win-x64.exe) |
-[Windows arm64](https://github.com/Fchat-Horizon/Horizon/releases/latest/download/F-Chat%20Horizon-win-arm64.exe) |
-[MacOS (Universal)](https://github.com/Fchat-Horizon/Horizon/releases/latest/download/F-Chat%20Horizon-mac-universal.dmg) |
-[Linux x64](https://github.com/Fchat-Horizon/Horizon/releases/latest/download/F-Chat%20Horizon-linux-x64.AppImage) |
-[Linux arm64](https://github.com/Fchat-Horizon/Horizon/releases/latest/download/F-Chat%20Horizon-linux-arm64.AppImage)
+[Windows x64](https://github.com/Fchat-Horizon/Horizon/releases/latest/download/F-Chat.Horizon-win-x64.exe) |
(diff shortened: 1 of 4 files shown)


FatCatClient:
There are _definitely_ some other places that allow you to enter character names as a raw input– like every single slash-command that lets you override an offline character's name by surrounding it with double quotes, but I'm not sure if there are any that'd run into trouble trying to find or create a character object for an empty name. We can do this to the website by entering ``/ignore ""`` though:

Untitled
_The greened out name is supposed to end up on the next line obviously._

Do you reckon it's doable for you to add a sanity check to those commands too for this PR?

I do agree with you that updating the Characters.State interface is a lot of work (32 references to the get() function if the Typescript LSP is to be believed), but I do think it's a good call to do this in the future. I sure wish we had Rust's Option\<T\> now :^)

FireUnderTheMountain:
the change in slash_commands should handle all commands that expect a Character.

FatCatClient:
That's on me for not looking at the change in the actual context instead of just looking at the diff from the PR page. My bad!

CodingWithAnxiety (approved):
Good quality changes. Approving.

Issue #119, opened 2025-04-26
1.30.2 seemingly reverts back to previous version. Maybe just visually?


Abydos2:
It is most likely just a visual thing (Hoping), but I was previously using 1.29.1. I hadn't updated to 1.30.1 because of the issue regarding text size and saw that there was a 1.30.2 release that had fixed the issue. So I downloaded it, ran the program and F-Chat said it was now on version 1.30.2. Was on there for a while, logged off. Next day when I log back in, the changelog website opened up like I had updated, checked the version and it said it was back to 1.29.1. Just to be on the safe side, I redownloaded the program, ran it again, updated to 1.30.2.. turned off f-chat, relaunched it and it was back to 1.29.1 again.

So far the steps seem to be:

1) Run an old version of F-chat Horizon.
2) Update to the latest version 1.30.2
3) Log off / shut down F-Chat
4) Start up / Log back in and it will go back to the previous version you had (At least visually I'm assuming)

I'm running windows 11 and did run the update program as an administrator, but seems to keep happening. I don't really know anyway to check if the chat is actually reverting to the old version before the update, or if its just giving the wrong version number. Clicking on it, redirects me to the old changelog page, only showing 1.30.1 and 1.29.1 ( https://github.com/Fchat-Horizon/Horizon/blob/main/CHANGELOG.md ).

FatCatClient:
The build system update for 1.30.x changed the app's ID (and name), which means that installing the newer version won't remove the old one. You'll have to click on uninstall for that one yourself.

![Image](https://github.com/user-attachments/assets/c32c7282-0b6e-4b01-a400-416cfdb310b0)
If you have both of these installed right now, try opening "F-Chat Horizon" instead of just just "F-Chat" and see if it opens the newer version. It should be under your list of all apps on the start menu, or in "Apps > Installed apps" in the settings app.

Abydos2:
I opened up my installed apps, and it does show both of them being installed like this:

![Image](https://github.com/user-attachments/assets/d0b8f7c1-eb18-41e8-b5b8-51e7cfaa9bfc)

Even though, I know I never installed 1.30.1, so I guess that is 1.30.2. While still having 1.29.1. I did see the thing you mentioned, so I  tried opening up the proper one (F-Chat Horizon) but it still says it is 1.29.1. Even opening up the normal / traditional f-chat one still says its 1.29.1

astrayblackcat:
[...]

PR #122, opened 2025-04-26
Fixes Windows notifications being stinky poopoo

### LINKED ISSUE #94: Notifications on Windows show the AppID rather than the name
Image

I think this might have something to do with the app ID being different in ``electron/package.json``— which is the one Squirrel gets when building the Electron app, versus the one defined in ``electron/main.ts`` [here](https://github.com/Fchat-Horizon/Horizon/blob/f798c39b57d067d55646e54c78cbb626bc33a9c4/electron/main.ts#L438).

~I'll add that I'm not using an official release, but instead built it from source myself. So this issue might be related to that.~ It seems to also happen if I use the official release.

FatCatClient:
After 1.30.1 the App ID for building in Electron was changed to ``net.flist.fchat``– except that it wasn't updated in main.ts. I don't really know how many people have already switched to post-1.29 versions, and I don't think flip-flopping the appId in electron's package.json is a good idea either. [We have already seen what happened because of the first change](https://github.com/Fchat-Horizon/Horizon/issues/119), so we better just stick with the new appId from now on.

Making this a draft PR because I want to see if this breaks something horribly first. It seems fine for now though. I also want to force push #123 (and the fixes for the issue below) in one commit, but I want to test the changes individually first. I don't yet know if changing the appID like this breaks stuff on existing Linux and Mac installations.

Fixes #94

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -435,7 +435,7 @@ function onReady(): void {
 
   log.info('Starting application.');
 
-  app.setAppUserModelId('com.squirrel.fchat.F-Chat');
+  app.setAppUserModelId('net.flist.fchat');
   app.on('open-file', createWindow);
 
   if (settings.version !== app.getVersion()) {
--- a/electron/notifications.ts
+++ b/electron/notifications.ts
@@ -28,7 +28,7 @@ export default class Notifications extends BaseNotifications {
           remote.getCurrentWebContents().id
         );
         conversation.show();
-        browserWindow.restore();
+        if (browserWindow.isMinimized()) browserWindow.restore();
         browserWindow.focus();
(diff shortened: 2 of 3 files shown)


FatCatClient:
Having tested this for a bit more, I at least know that it _does_ fix the notifications having the wrong name and doing absolutely nothing when you click them.
The downside is that when you click a notification it doesn't actually bring the app back into focus. It _does_ open the correct conversation (at least with one tab open), but...


image

Since switching tabs back and forth fixes this, **_and it works perfectly fine if the app window is just behind something instead of minimized_**, I'm going on a limb here and say that merging #89 would fix this too.

PR #128, opened 2025-04-30
Implement proxy settings

dupontcc:
Being able to set a proxy is probably not just useful to me so I thought this could be useful to others.
I think this should catch everything flist related and there is no earlier call to flist.

Currently this proxy would be shared for all flist tabs (since they all share a single context from what I saw). That makes sense but it may be a bit unintuitive ux wise.


FatCatClient:
Seems like a useful feature that was long overdue. Will check this out more intently after the current release is done.

CodingWithAnxiety:
Tested. Works well, Will merge into dev and make it into next (likely beta) release

PR #134, opened 2025-05-03
docs(readme): fix table of contents link to AUR package info

astrayblackcat:
This wasn't properly updated when it was switched to an AUR package, so just a small correction.

P.S. wish I could contribute more, been super busy :cry:


FatCatClient:
I wouldn’t sweat it too much. We’re all just volunteering here, and at this rate we’re still pumping out patches and improvements faster than Rising ever was.

PR #135, opened 2025-05-03
Experiment: Changes most F-Chat (Horizon) instances to be consistent

FatCatClient:
\>Website refers to it as just "Horizon"
\>Look inside
\>Half the code calls it "F-Chat Horizon" in places visible to users
\>The other half just calls it "F-Chat" in places visible to users

image

Screenshot 2025-05-03 at 20 07 35

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Table of Contents <!-- omit in toc -->
 
 - [Download](#download)
-- [F-Chat Horizon](#f-chat-horizon)
+- [Horizon](#f-chat-horizon)
   - [Features](#features)
   - [Goals](#goals)
 - [Installing](#installing)
@@ -27,7 +27,7 @@
 [Linux x64](https://github.com/Fchat-Horizon/Horizon/releases/latest/download/F-Chat.Horizon-linux-x86_64.AppImage) |
 [Linux arm64](https://github.com/Fchat-Horizon/Horizon/releases/latest/download/F-Chat.Horizon-linux-arm64.AppImage)
 
-# F-Chat Horizon
+# Horizon
 
 This repository contains a continuation of the heavily customized F-Chat Rising, a version of the mainline F-Chat 3.0 client.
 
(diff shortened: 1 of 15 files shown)


BootsieWootsie:
I think this is a good change. I think it's a solid one for making the client feel more consistent and unified.

The only place I think it might be contested is on the login screen specifically since you're connecting to F-Chat rather than Horizon there but other than that, I'm all for this change. Even then, I don't feel all that strongly about that specific location and I do have to admit it makes things look more consistent there too.

freenutsxd:
I agree that replacing the F-Chat titling with Horizon would probably be better for consistency, but we'd also have to think about what to rename within the Electron menus, whether we still want to call it "F-Chat Horizon" for transparency or simply "Horizon".

On a similar note, how would we feel about replacing the icon to something more unique and identifiable as _Horizon_ as well as opposed to the current reskinned F-Chat icon? 

Personally I don't mind either way, just something to consider. For reference, I believe XarChat uses it's own icon while Frolic keeps the default F-Chat icon.

FatCatClient:
Personally I’m in favour of full consistency and having every instance of the name just be “Horizon”— we can always just describe it as an F-chat client in the about screen and on the website. 

As for the icon, I had some ideas myself that I’ve floated around CWA in private, but haven’t settled on anything specific yet. Any time I’ve spent on this project recently was more so spent on new releases. 

I’ll have something new soon.

CodingWithAnxiety:
> Personally I’m in favour of full consistency and having every instance of the name just be “Horizon”— we can always just describe it as an F-chat client in the about screen and on the website. 
> 
> As for the icon, I had some ideas myself that I’ve floated around CWA in private, but haven’t settled on anything specific yet. Any time I’ve spent on this project recently was more so spent on new releases. 
> 
> I’ll have something new soon. 

CWA is a cool nickname. I like that. 

I'll get you an answer on the icon soon.

PR #139, opened 2025-05-04
1.31.0 Release

FatCatClient:
For the development branch PR, see #120.

[Changelog.](https://github.com/Fchat-Horizon/Horizon/blob/d2c5651bc7de57ee413eed0992ea23f4637c19dd/CHANGELOG.md)

[Pre-release builds.](https://github.com/Fchat-Horizon/Horizon/releases/tag/v1.31.0-beta2)

## Bookkeeping
Closes https://github.com/Fchat-Horizon/Horizon/issues/115
Closes https://github.com/Fchat-Horizon/Horizon/issues/94
Closes https://github.com/Fchat-Horizon/Horizon/issues/93
Closes https://github.com/Fchat-Horizon/Horizon/issues/34
Closes https://github.com/Fchat-Horizon/Horizon/issues/47
Closes https://github.com/Fchat-Horizon/Horizon/issues/37

PR #140, opened 2025-05-05
enhancement: Original colour option for gender markers

BootsieWootsie:
Implements the original gender-dots original colour as an option for the new gender icons as requested in https://github.com/Fchat-Horizon/Horizon/pull/139. Option disables whenever the markers are off or whenever the custom name colours are off to indicate that it requires these to change anything.

This version doesn't autoupdate, currently haven't found a good solution to make the colour change for this autoupdate. Easiest manual update is turning the icons off and on again. 

Might come back to see about a better method for this but the implementation of it currently doesn't align well with any of the attempted methods for the markers automatic updating, even the original method that was potentially poor on the performance. Ideal method performance-wise would just be watching for the setting changing in SettingsView itself and toggling markers off and back on again. I couldn't get it going in the little test run I had setting this up though.

This also includes a tiny fix for the new offline colour changing option, when using bookmark names they will properly go grey now when the bookmark goes offline.

Prettier has been run on it.

### DIFF
--- a/chat/common.ts
+++ b/chat/common.ts
@@ -78,6 +78,7 @@ export class Settings implements ISettings {
   risingShowHighQualityPortraits = true;
   horizonShowCustomCharacterColors = true;
   horizonShowGenderMarker = false;
+  horizonGenderMarkerOrigColor = false;
   horizonChangeOfflineColor = false;
   horizonNotifyFriendSignIn = false;
 
--- a/chat/SettingsView.vue
+++ b/chat/SettingsView.vue
@@ -354,6 +354,21 @@
       </div>
 
       <div class="form-group">
+        <label class="control-label" for="horizonGenderMarkerOrigColor">
+          <input
+            type="checkbox"
+            id="horizonGenderMarkerOrigColor"
+            v-model="horizonGenderMarkerOrigColor"
+            :disabled="
+              !horizonShowGenderMarker || !horizonShowCustomCharacterColors
+            "
+          />
+          Make gender icon use the original gender color instead of the custom
+          name color
+        </label>
+      </div>
+
+      <div class="form-group">
         <label class="control-label" for="horizonChangeOfflineColor">
           <input
             type="checkbox"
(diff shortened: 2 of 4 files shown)


chat/common.ts
```diff
@@ -78,6 +78,7 @@ export class Settings implements ISettings {
   risingShowHighQualityPortraits = true;
   horizonShowCustomCharacterColors = true;
   horizonShowGenderMarker = false;
+  horizonGenderMarkerOrigColor = false;
```
FatCatClient, line 81: This is not a rhetorical question, but do you think it's better to have this be disabled or enabled by default?

Personally I'd have made this be on by default, but I'm curious if that's just because I'd put my own preferences first.

FatCatClient:
>Might come back to see about a better method for this but the implementation of it currently doesn't align well with any of the attempted methods for the markers automatic updating, even the original method that was potentially poor on the performance. Ideal method performance-wise would just be watching for the setting changing in SettingsView itself and toggling markers off and back on again. I couldn't get it going in the little test run I had setting this up though.

This is actually broken for the "Show friends/ bookmarks in a different color" setting too– and if I'm going to guess it's been like that sometime after custom name colours were introduced. It's definitely not a thing in 3.0, but I can't be bothered to test if it's a problem in pre-Horizon versions right now.

If you want to fix that, that'd be cool. But I'm 99% sure it has the same cause as the thing you were struggling with and it's not something you introduced, so you're better off doing that in a separate pull request. And even then, I wouldn't lose any sleep over it. Something to do on a rainy Sunday afternoon perhaps 😉

chat/common.ts
BootsieWootsie: I personally think it's better to default it to disabled, it doesn't really convey any new information once you have the markers on and I personally still think it's better with this option disabled. Just looks nicer IMO. The other argument for leaving it disabled is just that the markers currently default to disabled too.

I do like having it as an option though, people think differently and have their own preferences for things so being able to choose is a good call.

PR #176, opened 2025-05-21
Font Awesome v6 Upgrade

### LINKED ISSUE #151: The "Show Ad Log" option has no icon
**Screenshots**

Image

0/10, asking for a refund

Just pick any of the [Font Awesome 5 icons](https://fontawesome.com/v5/search?ic=free) that look fitting, like ``fa-list`` or ``fa-ad``.

- OS: All
- Version: 1.31.0 (but really any version prior to Horizon)
- Theme: Any

freenutsxd:
Considering this a big 'maybe' PR since the style of the fonts changed in a way that some might not like. The upgrade itself was pretty straightforward and all the existing icons seem to work without any changes. It Just Works™️

In the screenshots below, the most notable change that might be contentious is the way they changed the ad icon. The text within the icon is smaller, so it's a little harder to read. It's little changes like this overall, but nothing _too_ jarring.

Left is v5, right is v6.
image

Also fixed #151 since the ad we were trying to use for the 'Show ad log' button was from the pro plan- we're only using the free plan. That commit can be merged without the v6 upgrade commit if we want to hold off.

### DIFF
--- a/chat/ChatView.vue
+++ b/chat/ChatView.vue
@@ -41,37 +41,37 @@
       </div>
       <div style="clear: both">
         <a href="#" @click.prevent="showSearch()" class="btn"
-          ><span class="fas fa-search"></span>
+          ><span class="fas fa-fw fa-search"></span>
           {{ l('characterSearch.open') }}</a
         >
       </div>
       <div>
         <a href="#" @click.prevent="showSettings()" class="btn"
-          ><span class="fas fa-cog"></span> {{ l('settings.open') }}</a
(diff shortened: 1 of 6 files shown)


FatCatClient:
I think the marginal differences are fine-- and the spacing for the top left buttons looks way better too.

Dunno why I thought FA7 was out already though :^)

FatCatClient:
I went through all menus pretty rigorously and I couldn't find a single broken icon. But apparently they added aliases for all the changed names anyway.

Issue #167, opened 2025-05-18
The app window title bar minimise button does not work when clicked on


CasperAsparagus:
**Describe the bug**
When the app is running left clicking on the minimise button (left most button on the top right 3 buttons - minimise, maximise/restore, close) has no effect. The app does not minimise but can be minimised / restored by left clicking the app icon on the KDE task manager (aka task bar in windows). 

**To Reproduce**

1. Load the application
2. Click the minimise button

**Expected behavior**
The application should minimise out of view into the task manager / task bar such that when the icon is clicked it is restored.

**Version and system info (please complete the following information):**

Distributor ID: Ubuntu
Description: Ubuntu 24.04.2 LTS
Release: 24.04
Codename: noble
Kernel: 6.8.0-60-generic
Architecture: x86_64
Desktop Environment: KDE Plasma
Display Server: x11
(I can provide more sys info on request)

FireUnderTheMountain:
Ancient problem. However, it only happens for me (openbox) when I'm not logged in; once login happens, it's gone. Can you confirm/deny it's the same for you?

CasperAsparagus:
Hi @FatCatClient, I'm using X11 rather than Wayland, and no I don't have any trouble with other applications that have custom title bars - Discord, Vivalid, and Deepin calculator all work with non-standard title bars. It's just this one app for some reason.

I tried out that PR also and still had no luck, same exact behaviour. Minimise wouldn't work logged in or not.

FatCatClient:
Just a quick shot in the dark, what happens when you press this Minimize option?

![Image](https://github.com/user-attachments/assets/f35cb0b0-4b50-43ac-b9bd-b5d01fa78997)



And since you mentioned using the AppImage version in your other issue, I'm just going to try and isolate as many problems as possible here, but have you experienced this problem with the .deb version, or even on a fork like [Frolic](https://github.com/Frolic-chat/Frolic) or the [original client](https://wiki.f-list.net/F-Chat_3.0/Installation)?

I'm also not ruling out on this being something horribly obscure with KDE either, especially since I know it's not standard with Ubuntu. Which exact version number are you running, if you can find that out?

CasperAsparagus:
Selecting the minimise option from the f-chat menu also has no effect either. It does do a little animated jerk as if it's about to minimise but then doesn't.

As for KDE versions this is what my about this system info says:
KDE Plasma Version 5.27.12
KDE Frameworks Version: 5.115.0
[...]

PR #192, opened 2025-05-29
Declare victor in mortal combat invoked during minimizing on linux - Fixes Fchat-Horizon/Horizon#167

FireUnderTheMountain:


FireUnderTheMountain:
Actually I do have one comment: Needs mac testing

FatCatClient:
I gave it a quick whirl and it seems to work fine there– unless there's a bunch of test cases that aren't implied but should be tested for. The Mac version uses the system's native window control buttons instead of our custom ones anyway.

If you could fix that Prettier check though, that'd be great.

FatCatClient:
I’m holding off on merging this for now since it’s mentioned in the comments for #167 that this does not fix the issue for them. If this isn’t an issue with their KDE install (which is my leading theory), my other guess is that it has something to do with non-native window control buttons being used. 

In that case, we’d instead need to configure the browser window object to use a titleBarOverlay on Linux instead. Maybe on Windows too, just to give it the mouse over functions that native window control buttons have.

PR #193, opened 2025-05-29
Switches scroll bars to Chrome native ones

FatCatClient:
This also sets a global CSS color scheme for all themes– in this case so that the scrollbar's gutter (which is only visible when you manually drag the thumb) respects whether or not the theme is light or dark, rather than consult the OS's color scheme setting, but in the future this will be useful if we do things like move certain dialog windows into global modal dialogs.

https://github.com/user-attachments/assets/99a8b0c9-3aaf-47b2-abdb-31f41e75405b

I wasn't a big fan of the custom ones we had because I felt they were too obnoxiously visible. These are a lot more subtle (and they're also more in line with the browser experience anyway).

### DIFF
--- a/scss/themes/_chat.scss
+++ b/scss/themes/_chat.scss
@@ -39,27 +39,3 @@
     --#{$varName}: #{$value};
   }
 }
-
-* {
-  &::-webkit-scrollbar-track {
-    box-shadow: inset 0 0 8px $card-border-color;
-    border-radius: 10px;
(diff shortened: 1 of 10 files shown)


FatCatClient:
Closing this because native scroll bars look like ass on Windows. I had assumed they’d look like the ones in Chrome or Edge on that platform. 

We’ll come up with a better alternative eventually.

Issue #209, opened 2025-06-10
The "Show bookmarks/ friends in a different color" setting doesn't work in themes that override the color


FatCatClient:
**Describe the problem**

![Image](https://github.com/user-attachments/assets/2025dc8f-a5c1-4bf2-bfad-360be6503383)

All this needs is for the color for that specific class to also be set for the Dracula theme.

**Version and system info (please complete the following information):**

- OS: Windows
- Version: 1.32.0-beta1
- Theme: Dracula

PR #210, opened 2025-06-11
fix: override bookmark and friend colors in Dracula theme

Keerthi421:
## Description:
This PR addresses Issue #209 by adding color overrides for the .user-bookmark and .character-friend/.friend-item classes in the Dracula theme. This ensures that bookmarks and friends display in the correct colors, consistent with the theme's palette.

## Changes Made:
Added color overrides in scss/themes/chat/dracula.scss for .user-bookmark and .character-friend/.friend-item classes.
Used $dracula-green for bookmarks and $dracula-cyan for friends to maintain theme consistency.

### DIFF
--- a/scss/themes/chat/dracula.scss
+++ b/scss/themes/chat/dracula.scss
@@ -94,8 +94,3 @@ $genders: (
 .message-event .user-bookmark {
   color: $dracula-green;
 }
-
-.character-friend,
-.friend-item {
-  color: $dracula-cyan !important;
-}


scss/themes/chat/dracula.scss
```diff
@@ -89,3 +89,13 @@ $genders: (
     --#{$varName}: #{$value};
   }
 }
+
+.user-bookmark,
+.message-event .user-bookmark {
+  color: $dracula-green !important;
```
FatCatClient, line 95: This doesn't have to be ``!important``. The theme should automatically override the base colours.

PR #212, opened 2025-06-11
Eicon Update fixes & adjustments

FireUnderTheMountain:
Briefly,
* Fix to the bug where randomly getting an eicon would cause it to be unsearchable
* Allow users to use spaces as part of their search query
* Error handling for Xariah responses that might have resulted in unshuffled or empty eicon store
* Error handling for failure to write to disk
* Stop trying to pull my vscode config into commits


FatCatClient:
If you want a VCCode .gitignore setup, you should probably look at [this.](https://www.toptal.com/developers/gitignore/api/visualstudiocode) There are technically _some_ files in that folder that we might want to version control (like a recommended plugins file), but I don't use VSCode myself so I don't really think I should be the one to set that up.

FireUnderTheMountain:
There's a [decent discussion](https://stackoverflow.com/questions/32964920/should-i-commit-the-vscode-folder-to-source-control) on stack overflow that has all the major points covered, but ultimately what it boils down to is whether or not there's anything worth **enforcing** for vscode users instead of everyone. There **can** be, but there's a few major points from the discussion:

* Even though `settings.json` is supposed to be "workspace" settings, it's really "workspace settings for this user" - same for extensions. I use an indent highlighter to deal with 2-space indents that really shouldn't be forced on other people who don't have the issues with it that I do. I also use different line-wrap indentation settings in this project than in others; another change that shouldn't be forced on other vscode users.
* I'm not aware of `tasks` or `launch` being needed for anything important in this project; otherwise there would be some kind of consideration for other editors as well. But it does seem as if that could be useful in the future.
* In the SO discussion, someone points out Git uses (used?) an init script to set up the environment for new contributors. This may be something worth doing in the future, and forwards the idea that the actual created config files should not be version controlled. A quote from the thread, from the Git developers:

   >This patch adds a script that helps set up the environment to work effectively with VS Code: simply run the Unix shell script contrib/vscode/init.sh, which creates the relevant files, and open the top level folder of Git's source code in VS Code.

I would recommend something like that in the future, if such a thing is ever desirable. That said, it may be a good idea to document some of this directly inside the .gitignore file. Thoughts on something like this?:

```
### VisualStudioCode ###
### Many of these are ignored on sound advice and not by experience;
### Because there is no enforced editor app, pushing settings and extensions on other vscode users doesn't make any sense.
### If there's valid reason to propogate any of the below files, file an issue and a discussion can be had.

.vscode/
#!.vscode/settings.json
#!.vscode/tasks.json
#!.vscode/launch.json
#!.vscode/extensions.json
#!.vscode/*.code-snippets

# Local History for Visual Studio Code
.history/

# Built Visual Studio Code Extensions
*.vsix

### VisualStudioCode Patch ###
# Ignore all local history of files
.history
.ionide
### End of VisualStudioCode ###
```

FatCatClient:
For the sake of the changelog, these are bugs that also happened with the Rising eiconstore, right? @FireUnderTheMountain :

- Allow users to use spaces as part of their search query
- Error handling for Xariah responses that might have resulted in unshuffled or empty eicon store

PR #225, opened 2025-06-17
Fixes icon bbcode linking to improperly capitalised profile

BootsieWootsie:
Quick fix for when user inputs an improperly capitalised version of the name of a profile when using the icon bbcode.
Fixes the profile preview opening to a version with the capitalisation they used at the top, presents the proper capitalisation. Also fixes the text on hover of icon too in the same circumstance.

### DIFF
--- a/bbcode/Editor.vue
+++ b/bbcode/Editor.vue
@@ -126,6 +126,7 @@
   import { default as IconView } from './IconView.vue';
   import { default as EIconSelector } from './EIconSelector.vue';
   import Modal from '../components/Modal.vue';
+  import { Character } from '../fchat';
 
   @Component({
     components: {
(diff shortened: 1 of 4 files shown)


FatCatClient:
I'm not quite sure what this is supposed to fix exactly? In cases where the floating mini preview displays the correct (cached) capitalisation for an offline character, typing a both an icon and a user tag with wrong caps still shows the incorrect capital letters being used.

It works fine for users that are (or have been) online while you're connected, or ones in your Recent Conversations list, but it always has afaik.

BootsieWootsie:
Ah, yeah, it would also do this for online/recent characters with the icon display specifically.

I had done this decently late and not realised the issue with this occuring either way with users who haven't been online/recent even with the fix.

I have a version in the back that fixes the mini avatar but it may be wiser to come at this from another angle and change the profile viewer itself's handling of the name.

PR #240, opened 2025-06-25
Add cache and disk store for draft messages

AriannaAltomare:
This PR adds a frequent in-memory cache and periodic on-disk copy for in-progress/draft messages in the client, and saves the results per-character. The intended user-visible behavioral improvements are as follows:

1. Accidentally closing a conversation will restore the text in the message box as it was when reopened.
2. Closing a character tab (without the full overall process closing) and later reopening it will restore all draft messages (saved to disk every 60s, can be adjusted by user in Settings > Horizon).
3. Reopening the program after the process closes (whether due to a crash or the like) will restore all draft messages (same disk saving frequency as above).

All three have been tested and currently work as-designed in an ideal use-case. Users can also disable/enable this behavior entirely in Settings > Horizon.

### DIFF
--- a/chat/conversations.ts
+++ b/chat/conversations.ts
@@ -27,6 +27,12 @@ import Bluebird from 'bluebird';
 import log from 'electron-log';
 import isChannel = Interfaces.isChannel;
 
+/**
+ * @constant
+ * How frequently the contents of the textbox in each conversation should save to the in-memory cache, if enabled.
+ */
+const CONVERSATION_CACHE_UPDATE_FREQ_IN_MS = 1000;
+
 function createMessage(
   this: any,
   type: MessageType,
@@ -970,6 +976,15 @@ async function testSmartFilterForChannel(
   return false;
 }
 
+/**
+ * Sets up the backend draft message cache if needed, then populates the input textbox with the saved draft if one exists. At regular
+ * intervals, this also saves the current draft message to the cache to be recovered in the event of a full tab crash, disconnect (in the
+ * event the tab does not attempt to save state), or process/system crash.
+ * @function
+ * @param {Conversation} this
+ * The conversation which will control the textbox and update interval. Upon close, the interval will be terminated to prevent leaks.
+ * @internal
+ */
 async function initConversationCache(this: Conversation): Promise<void> {
   // Restore message draft if it exists (e.g. accidentally closing the tab). Be sure the cache is reset for a new character if needed.
   await core.cache.conversationDraftCache.resetCacheIfNeeded();
@@ -996,7 +1011,7 @@ async function initConversationCache(this: Conversation): Promise<void> {
       this.enteredText
         ? core.cache.registerConversationDraft(this.name, this.enteredText)
         : core.cache.deregisterConversationDraft(this.name);
-    }, 1000);
+    }, CONVERSATION_CACHE_UPDATE_FREQ_IN_MS);
   }
 }
 
--- a/learn/conversation-draft-cache.ts
+++ b/learn/conversation-draft-cache.ts
@@ -1,6 +1,18 @@
+/**
+ * @module conversation-draft-cache
+ * Maintains an in-memory cache of draft messages, and occasionally saves them to disk.
+ */
+
 import { Cache } from './cache';
 import { getDrafts, saveDrafts } from '../electron/filesystem';
 import core from '../chat/core';
+
+/**
+ * @constant
+ * The minimum wait time between doing disk saves. As the disk save currently uses fs.writeFileSync, this is primarily to protect
+ * against the event loop blocking due to rapid writes.
+ */
+const MIN_CACHE_DISK_SAVE_IN_SECONDS = 5;
 
 export interface ConversationCachedMessage {
   channel: string;
@@ -11,6 +23,15 @@ export class ConversationDraftRecord {
   channel: string;
   message: string;
 
+  /**
+   * Fundamental layout of an F-Chat message. Sender is currently implied by core.connection.character.
+   * @function
+   * @param {string} channel
+   * The intended recipient of the message, either a character name or a channel name.
+   * @param {string} message
+   * The draft text as it currently exists in the input textbox.
+   * @internal
+   */
   constructor(channel: string, message?: string) {
     this.channel = channel;
     this.message = message || '';
@@ -26,6 +47,12 @@ export class ConversationDraftCache extends Cache<ConversationDraftRecord> {
   private currentlyCachedCharacter = '';
   private resetListenerActive = false;
 
+  /**
+   * Initialize the cache, including pulling the backup from disk if it exists. If setting.horizonCacheDraftMessages is false (opt-out)
+   * then the cache will simply be ignored. Otherwise, this feature is on by default.
+   * @function
+   * @internal
+   */
   async loadCache(): Promise<void> {
     if (!this.resetListenerActive) {
       core.connection.onEvent('connected', () => {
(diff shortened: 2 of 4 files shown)


AriannaAltomare:
Absolutely, no rush on this - fortunately most of my TODOs are edge cases and preventing any Bad Things™ from file writes or anything overwriting what someone's in the middle of typing, so hopefully those will be fairly straightforward.

The settings should be pretty straightforward fortunately, but I agree that any of the attached UI work can wait until after the Vue fun! As for the purging behavior, I'll think on that a little more. Admittedly I'm from a long career line of SRE and cybersecurity, so UX and design isn't my forte - I'll ask around a bit and see if there's a good idea here. Off the cuff, I could imagine anything from keeping it "forever" in a Drafts menu to just having a small modal asking if you want to restore things (kinda like Chrome?) At the very least, it's a minor decision in the scheme of things.

I'll reach out once I feel that the backend is rock solid and that there aren't any performance issues, then we can follow up on any of the design post-UI rework! 🙂

learn/conversation-draft-cache.ts
```diff
@@ -0,0 +1,66 @@
...
+    delete this.cache[k];
+
+    // TODO: Right now this also runs if people dump their message box (e.g. Ctrl-X to hold a message).
+    // Intent is just to run on message send, maybe just reveal a method for that? deregisterAndClear?
+    // Otherwise, it just leaves empty objects in the disk file for every char/channel they interact with, which is ugly.
+    this.saveCacheToDisk();
+  }
+
+  private saveCacheToDisk(): void {
+    // Buffer close writes. Missing an occasional save isn't the end of the world.
```
FireUnderTheMountain, line 59: If you're worried about writes overlapping, using [an async write with an AbortController](https://stackoverflow.com/questions/74529131/node-js-how-to-cancel-a-writefile-operation) handled by the draft class may be a solution.

chat/conversations.ts
```diff
@@ -461,6 +481,11 @@ class ChannelConversation
   set enteredText(value: string) {
     if (this.isSendingAds) this.adEnteredText = value;
     else this.chatEnteredText = value;
+
+    // FIXME: Saving on every edit, consider a slight buffer if non-performant.
```
FireUnderTheMountain, line 485: I don't think this will cause performance issues (but testing could prove me wrong) but if you end up needing to do any complicated logic (like only saving if user has 5+ unsaved messages, or total unsaved message length is gratuitous) you can use a debounce with a maxWait to push all that off to some later time and only run it every handful of seconds.

AriannaAltomare: Agreed it'll mostly likely be non-problematic, but for now I've set it to just save every couple seconds to try and be kind to those on genuinely poor hardware. Should be unlikely (I hope) that someone manages to write a whole essay and then loses the tab in that short timeframe (even with the 60s default disk backup writes, unless they change that setting).

learn/conversation-draft-cache.ts
AriannaAltomare: Incidentally, after I looked a little deeper, the under-the-hood "writeFile" helper for the logs actually uses fs.writeFileSync - so we should be good there!

AriannaAltomare:
@FatCatClient Got a few hours this weekend to clean up the PR and add in some attached settings, should be good for a review and more firm feedback! My own testing seems like it works appropriately, though the new settings require a save and reset for them to take effect (FWIW, this is very minimal atm as they're in the per-character settings under the Horizon tab - it's highly likely that's not where their permanent home should be, especially after the Vue 3 release).

FatCatClient:
I'll have a proper look at this somewhere today or tomorrow. 
> (FWIW, this is very minimal atm as they're in the per-character settings under the Horizon tab - it's highly likely that's not where their permanent home should be, especially after the Vue 3 release).
We might keep it as a character specific setting, even after we move most of those as a global one instead. We'll see!

mobile/chat.ts
FatCatClient, line 1: You really do not have to format and commit the entirety of the mobile directory or the pnpm-lock files though.

AriannaAltomare: Aha, found it - looks like some automation ran it from within /electron which made it ignore the file. Should be back to normal now.

FatCatClient: That wouldn't happen to be from either a38c9cb3aaa47f250a0b01f64294c0c4d2045e51 or 7f4966687f8c3b05d410c923b00c2b4ebf5242f7, right? I was certain that it only runs the formatter on files you actually staged (and even then would have to ignore the stuff in .prettierignore).

AriannaAltomare: And I think I see my second mistake - I rebased off of main instead of development. I bet there was some code diff there that staged those files and, with the .prettierignore not working from the subdirectory, ran it.

chat/SettingsView.vue
```diff
@@ -276,6 +276,35 @@
...
+      <h5>Draft Messages</h5>
+
+      <div class="form-group">
+        <label class="control-label" for="horizonCacheDraftMessages">
+          <input
+            type="checkbox"
+            id="horizonCacheDraftMessages"
+            v-model="horizonCacheDraftMessages"
+          />
+          Automatically save and restore in-progress messages while using
```
FatCatClient, line 288: Mind converting these two descriptions (and the title) to use the localization system instead? We'll be doing this across the board later on, so we might as well not increase the workload for the future.

FatCatClient:
Gave it a shot briefly and couldn't actively break it myself. Looks pretty good so far, so I just have a couple of minor, nitpicky things in mind.

learn/conversation-draft-cache.ts
```diff
@@ -0,0 +1,112 @@
...
+    delete this.cache[k];
+    this.saveCacheToDisk();
+  }
+
+  private saveCacheToDisk(): void {
+    if (!this.useCache || !core.connection.character) return;
+
+    // Buffer close writes. Missing an occasional save isn't the end of the world.
+    const now = Date.now();
+    if (now - this.lastCacheSave < 5000) return;
```
FatCatClient, line 107: Magic number. ``1000 * MIN_CACHE_SAVE_IN_SECONDS`` or however you want to call it 😉

learn/conversation-draft-cache.ts
```diff
@@ -0,0 +1,112 @@
...
+      settings?.horizonSaveDraftMessagesToDiskTimer ||
+      this.diskSaveTimerInSeconds;
+
+    if (!this.useCache) {
+      this.cacheAlreadyLoaded = true;
+      return;
+    }
+
+    // Check possible accidents from the user (mostly a backup check in case the setting somehow gets set to e.g. 0)
+    if (this.diskSaveTimerInSeconds < 5) this.diskSaveTimerInSeconds = 5;
```
FatCatClient, line 53: It's a raw number in the settings menu too, but I'm fine w/ that for now (unless we want to go and make a generic ``consts/settings.ts`` file or something). But for now it'd be a tiny bit cleaner if this 5 were a const in this specific file too.

chat/conversations.ts
```diff
@@ -524,6 +535,7 @@ class ChannelConversation
 
   close(): void {
     core.connection.send('LCH', { channel: this.channel.id });
+    clearInterval(this.cacheInterval);
```
FatCatClient, line 538: Was it a design decision or an oversight to not clear the conversation cache here too? (same with close on line 673).

Not a right or wrong answer, but since this is done directly through user input I was wondering if you thought it should keep the cached message or not.

FatCatClient:
Also I forgot to mention it in my review comments, but if you have the time to, could you add some JSDoc/ TypeDoc-style comments to your new methods and files? It'd save us some time in the future for when we want to export the TypeDoc stuff to the website's docs.

Don't have to overdo it either, you can just do it like [here](https://github.com/Fchat-Horizon/Horizon/commit/a4ec493c85b155bbd70e2db160b96f6ad95bb49a). And if you're really lazy, VSCode's auto complete is actually really good at writing those.

chat/conversations.ts
AriannaAltomare: Yeah, it was deliberate in this case but it's really a judgement call. My thought is to times where I've accidentally closed a tab while trying to click elsewhere, and ended up losing some work. FWIW, it would already reload the last save from disk, unless we wanted this to also deregister the cache (which is an option!)

AriannaAltomare:
@FatCatClient Done and done, feel free to adjust anything text-wise in the formatting of the comments, but I did my best to adhere to your style in the reference. 🙂

PR #241, opened 2025-06-25
EIcon Store episode 3: Return of the Robust Structure Checking

FireUnderTheMountain:
Implements:
* A more robust store check to avoid saving a bad store or to redownload the whole store in case of irregularity
* Privitize members that shouldn't be accessed publicly
* TSDoc notes - In the vscode tooltip, they display fine - but they may need additional formatting depending on what their final form is supposed to be. To know how to adjust them further, I'd have to see them displayed in the intended format.
* Move the hardcoded update check to the top of the page for easier reference

### DIFF
--- a/learn/eicon/store.ts
+++ b/learn/eicon/store.ts
@@ -108,7 +108,7 @@ export class EIconStore {
    * error is caught, logged, and ignored.
    */
   private async save(): Promise<void> {
-    if (this.lookup.length) {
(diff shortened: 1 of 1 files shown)


FatCatClient:
Tested with a bunch of corrupt example files I had, and it seems to work just fine.

We'll look at whether or not the TSDocs need some touching up after we've actually gotten TypeDoc to work with the Vue files.

PR #244, opened 2025-06-27
Add a QuickJump menu to easier keyboard navigation

CrimsonLotusCalamity:
For me F-Chat has a major problem: The lack of modern keyboard navigation. In this **very rough** draft of a PR, I quickly threw together a QuickJump menu that let's you jump to chats and channels! Use `CTRL + SHIFT + P` to access it.

image

Let's discuss an actual proper implementation of such a feature!


CrimsonLotusCalamity:
@FatCatClient I have to admit that this is the first time coding in vue. The background is blurred so it looks a lot better with other stuff behind it.

Think of this as a Proof of Concept. I honestly do not know if I am competent enough make worthwhile contributions, but I am currently using this as my personal fork and it really adds so much QoL.

CrimsonLotusCalamity:
As said, first time working with Vue and TS is not really my forte anyways!

chat/ConversationView.vue
FatCatClient, line 1: Any particular reason this has to be added to both components? Does calling it while typing in a conversation intercept the call from the parent component?

electron/package.json
FatCatClient, line 1: This dependency doesn't have to be in both the electron and the main package.json. If it happened because of a git pre-commit hook, you'd have to run the ``git commit`` command from the root directory.

electron/pnpm-lock.yaml
FatCatClient, line 1: See package.json

chat/QuickJump.vue
```diff
@@ -0,0 +1,459 @@
...
+    onVisibilityChange(newValue: boolean): void {
+      if (newValue) {
+        // Update results when showing
+        this.updateResults();
+      }
+    }
+  }
+</script>
+
+<style lang="scss" scoped>
```
FatCatClient, line 305: I _love_ the look for this, but it really doesn't fit in with the rest of the Bootstrap-styled components. I don't think we should change it now, but if other people are embracing this look too we'll have to consider generalising some of the CSS here.

chat/QuickJump.vue
```diff
@@ -0,0 +1,459 @@
...
+          display: flex;
+          align-items: center;
+          gap: 12px;
+          cursor: pointer;
+          border-bottom: 1px solid rgba(255, 255, 255, 0.1);
+          transition: all 0.2s ease;
+          position: relative;
+
+          &:hover {
+            background-color: rgba(59, 130, 246, 0.1);
```
FatCatClient, line 371: Preferably this is a color based on the theme too, could always just use a SASS function to get a more subtle version of the primary color.

chat/QuickJump.vue
```diff
@@ -0,0 +1,459 @@
...
+            color: var(--textColor);
+          }
+
+          &.selected {
+            background-color: var(--primaryColor);
```
FatCatClient, line 376: I don't think these colours seem to apply at all. The list items only get a background color if you hover over them and then move the selection with your arrow keys.

For theme colours you'll want to use ``border-color: var(--primary);`` or some SASS function variation of that.

chat/QuickJump.vue
```diff
@@ -0,0 +1,459 @@
...
+        });
+      }
+
+      // Sort by recent activity (unread first, then by last message time)
+      results.sort((a, b) => {
+        if (!a.conversation || !b.conversation) return 0;
+
+        // Prioritize conversations with unread messages
+        if (
+          a.conversation.unread !== Conversation.UnreadState.None &&
```
FatCatClient, line 164: This makes a lot of sense to do. Maybe it'd be even better if the search result also showed why certain conversations are being prioritised. Could be something as simple as a badge (which has a bootstrap class already 😉).

chat/QuickJump.vue
```diff
@@ -0,0 +1,459 @@
...
+          class="quick-jump-new-conversation"
+          :class="{ selected: selectedIndex === -1 }"
+          @click="openNewConversation"
+          @mouseenter="selectedIndex = -1"
+        >
+          <span class="result-icon">
+            <span class="fas fa-plus"></span>
+          </span>
+          <span class="result-name">{{
+            l('quickJump.openNewConversation', searchQuery)
```
FatCatClient, line 56: Probably a good idea to make this look similar to the selected items, that way people will know that pressing Enter works here too.

PR #258, opened 2025-07-07
Fixes duplicate icons for gender icons after FontAwesome update

BootsieWootsie:
Transgender and Shemale have the same icon currently, FontAwesome 6 renamed transgender to mars-and-venus and transgender-alt to transgender. 

It didn't inherently break things due to some backwards compatibility allowing transgender-alt to still load but as transgender instead.

This is a quick fix for that, tested and functional!


BootsieWootsie:
Here you go!
 For current release build
 For this PR

PR #259, opened 2025-07-07
Fixes a tiny capitalisation bug cause I can't unsee it (Recent Conversations)

BootsieWootsie:
Recent conversations does not match any of the other capitalisation, it's fixed. This is a single letter change, it is worth it.


FatCatClient:
Thank the gods. Horizon is saved 

image

FatCatClient:
Are we going to keep this PR up or were you going to check the consistency across the board first?

BootsieWootsie:
I'd say just keep this one as is since it does what it says on the tin and I'll work on something bigger (with localisation inclusions too)

FatCatClient:
Do you think those big changes are something you'll be doing within the next week or so? That way I know if I have to wait for it for the next update or not.

FatCatClient:
I think every case of text not being handled by the localisation files is from something added after 3.0, so that’s significantly narrow the search.

Issue #144, opened 2025-05-09
[hr] tags inside certain other tags don't render


FatCatClient:
**To Reproduce**

BBCode:

```bbcode
[hr]

[sub]
sup
[hr]
[/sub]

[sup]
sup
[hr]
[/sup]

[big]
[hr]
big
[/big]

[small]
[hr]
small
[/small]

[heading]
[hr]
heading
[/heading]


[b]
[hr]
[/b]

```

**Screenshots**

![Image](https://github.com/user-attachments/assets/afc13ea4-96ff-4b19-9734-ea7b22d87b9b)
_On the website_

![Image](https://github.com/user-attachments/assets/b68b18b1-8390-4d33-b044-9c4f8fee4f8b)
_In the profile viewer_

**Version and system info (please complete the following information):**

- OS: Platform independent
- Version 1.31.0-beta2 (but likely older)

FatCatClient:
The website's BBCode parser is weird like that. If I remember, you can't put [color] tags inside a [sub] or [sup] tag either, but you _can_ put them inside a color tag. Even if logically, this wouldn't break anything.

So if you want individual words inside the subscript to have a color, you have to do this weird ``[color=red][sub]foo[/sub][/color][sub]bar[/sub]`` construction.

PR #261, opened 2025-07-07
Bunch of profile viewer BBCode fixes

BootsieWootsie:
Fixes a good few things.

- Color tags that were done incorrectly did not parse in profile viewer but did on website
- HR tags would often break if near other things or inside certain other bits of bbcode
- sub has been broken in profile viewer (it's barely distinguishable from normal text)

All these are fixed, viewable profiles should go up a lot. This does not account for incorrect color bbcode in chat to keep parity with the main client.


FatCatClient:
Before I test, do the [hr] fixes fix the stuff in #144? For tracking the issues to be closed with the PR to main ;)

BootsieWootsie:
Correction, it does fix it inside bold. So yes, should completely close that!

PR #275, opened 2025-07-12
Quick fix to make inline data properly update in profile viewer when cache expires

BootsieWootsie:
Just slots in the line to update the available inlines into the refreshCharacter() function from load(), shouldn't cause issues since all the same data in there should be available. 

Refresh just makes sure it comes new rather than from the cache and this makes sure that includes the inline data it already collects is just stored properly.


BootsieWootsie:
Added another fix into this, fixes eicons messing up in headers too. One day I'll actually get all of these!

PR #281, opened 2025-07-14
Inlines now don't render in profile viewer if it cannot find an actual inline

BootsieWootsie:
This is also site functionality that's not mirrored by the client so far, it makes both of the other inline errors blank too. This iteration is actually somewhat inconsistent with the site since guestbook shows erroring inlines on the site but will now not show them in the client when they are broken. IMO, that's not too big of a deal.

This one I'll say is a maybe to add since inlines not being found can be a good indicator you need to refresh profiles and this only really matters if someone deletes an inline on the site but doesn't remove the bbcode.


FatCatClient:
Yeah, tbqh it might actually kind of help to see an inline being visibly broken through this.

PR #284, opened 2025-07-14
1.32.3 Release

### LINKED ISSUE #5: Connectivity issues.
This is something I kept asking of Rising for a few years, but since there's a full-fledged team, I thought maybe I could put this out there. Unlike the website or the original app, if you lose connection, it'll recover and keep your text unless there's a connection error and you need to be relogged. Likewise, the F-chat Rising app just deletes your text period, even if you regain connection a second later. As someone with spotty internet, this is a **massive** headache. If you guys could add a feature to save it to your clipboard or to keep it period in a log in case you have to relog, it'd be **massively** appreciated!

### LINKED ISSUE #121: Changelog should display inside the app
**Is your feature request related to a problem? Please describe.**
I think it's kind of annoying that installing a different version forces a tab in your browser to open, a tab that also just shows the most recent changelog-- which isn't necessarily the version you just updated to.

**Describe the solution you'd like**
Just make it a pop up window like the about screen. Or something like the way other apps show available updates:

Image

Could easily just add the CHANGELOG.md file to the build and then have a markdown parser show it as HTML. Either as part of the build process (which has my preference), or on the fly.

### LINKED ISSUE #144: [hr] tags inside certain other tags don't render

### LINKED ISSUE #256: Feature Request: Make "update available" notification BG color disappear after clicking the menu
**Is your feature request related to a problem? Please describe.**

When a new version of Horizon is released, the main menu button/element with the gear icon gets a bright BG color, similar to a tab with a new message, mention, etc. 

Image

This is a good way to notify the user of an update being available, but it is a major distraction as **a persistent thing.** I only need to see it once and should be allowed to decide if and when I want to take action without being reminded of it _**all the time.**_ The bright BG's should be reserved for things I haven't seen yet, so I'm sure to notice them.

**Describe the solution you'd like**

Like every new "unread message" notifier, it should disappear after the user has viewed the message. In this case, opening the menu – i.e. clicking the button with the gear icon – could count as having viewed it. 

Alternatively, although less preferably, the notification could be something much subtler that doesn't get mixed up with actual new messages, even at a glance. Personally, I'd be hesitant to use the attention color with a persistent notifier at all, and would opt for an extra icon or a small dot of a neutral color, instead – or something like that. I wouldn't see this as a permanent solution, following the reasoning above, but it might be better than the current, somewhat overbearing implementation.

A third potential option could be a notification similar to the ones you get for notes and such, in the bottom right corner. I think the current position is probably better, though, due to its direct association with the main menu that has the link to the update and changelog.

If I had to choose between having no notification at all (the menu item aside) and the persisting, distracting one like now, I'd pick the former. Design-and-usability-wise, the visuals shouldn't signify "new and unread" when that simply isn't the case after first viewing. 

The extra menu item for the update is also persistent (as it should be), so the info still remains available to the user until they _are_ ready to take action, regardless.

---

Does the client currently simply check if there's a version newer than the one you're using available and "nag" about it until there isn't (because you're up-to-date again)? 

I suppose the feature would have to take into account that the user may skip versions but still wants to get notified of all new releases like this, without the notifications being a bother. This leads me to think that these notifications _should_ perhaps follow the same logic as any other unread messages that become "read" after viewing – as described above – whether or not the technical or design solution for them is similar to "normal" notifications (which seems unlikely). Obviously, this may complicate the feature some from its current implementation, depending on what that implementation actually is.

### LINKED ISSUE #276: [1.32.3-dev3] User icons are missing on the profile on sidebar, eicons appear to be too.
**Describe the bug**
Sidebar profile does not render any IconView or Eicon elements in current dev build. I suspect it's my fault but I'm not wholly sure _why_ other than some odd parser issue cause it should just inherit from bbcode/standard and the sidebar preview is just character_page stuffed into the sidebar. I'm looking at some fixes for it right now so will keep you posted on that.

**To Reproduce**
Go to any profile with icons and/or eicons visible in the regular profile viewer, they will not render in the sidebar version of the profile.

**Expected behavior**
Icons and eicons should appear.

- OS: Windows 11
- Version: 1.32.3-dev3

FatCatClient:
Closes #5
Closes #121
Fixes #144 
Closes #256 
Fixes #276


chat/SettingsView.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <modal
-    :action="l('settings.action')"
+    :action="l('settings.character')"
```
AriannaAltomare, line 3: Visual nit I just noticed in the beta build: Currently this makes the "Change Settings" confirmation button read "Character Settings" instead, think this should remain .action (or have a new localization for "Save Character Settings")

AriannaAltomare: image

CodingWithAnxiety:
Do you think we should wait for RPM support before merging this? Or leave RPMs in dev for the time being and add it in 1.33 as the 'big change', since that'd encapsulate an entire CI/CD change to use the new build script.

PR #285, opened 2025-07-15
Formatting parity for bbcode in profiles

little-voice:
I noticed a few visual discrepancies with how tags that change the text size behave in the profile viewer VS F-list. The attached changes aim to fix these.

- Allow `[small]` and `[sup]` tags within `[heading]`s
- Use the same `font-size` value for `[big]` and `[small]` as the site. These are expressed in `em` since they're supposed to scale with the parent component (namely `[heading]`)
- Update the styles for `h2` elements within bbcode components (`heading` gets translated to `h2` in Horizon's implementation) to match the site's
- Tweak styles for `sup` and `sub` elements only inside the profile viewer. There are minor differences with how these get rendered in the chat VS on profiles.

I created the **Vitaly Carapathian** test profile to verify the changes for various valid tag combinations. I'll attach screenshots here as well.

image

*Website, Windows/Chrome, Default theme*

image

*Horizon current stable release, default theme*

image

*Dev build for this branch, default theme*

### DIFF
--- a/bbcode/standard.ts
+++ b/bbcode/standard.ts
@@ -118,7 +118,9 @@ export class StandardBBCodeParser extends CoreBBCodeParser {
           'color',
           's',
           'big',
+          'small',
           'sub',
+          'sup',
           'hr',
           'img',
           'eicon'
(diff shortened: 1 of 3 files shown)


FatCatClient:
Looks good to me. Thanks for that (and giving a profile to quickly review the changes on)!

PR #294, opened 2025-07-18
Refactor character state management

### LINKED ISSUE #143: Userview right click menu is (partially) case sensitive
**Describe the bug**
This is a super specific and minor bug, but in cases where a UserMenu component is opened from a UserView element while the UserView element is created with a profile name that has different casing from the real one, the bookmark button always assumed you do not have the user bookmarked even if you actually do.

**To Reproduce**

1. Have user X bookmarked while they're not online
2. Write their name with incorrect casing in a [user] tag
3. Right click the UserView element created when you send or preview the message
4. The bookmark button doesn't recognise you have the user bookmarked, but clicking the bookmark button returns a server error about having already bookmarked them.

You can also do this by opening a private convo with a name that uses incorrect casing, where it will also happen for the names in the Recent Conversations menu. But once somebody logs in, it uses the cached casing for all UserView elements, as far as I know

- OS: All
- Version 1.30.3, but I think this is a 3.0 era bug.

little-voice:
Fixes #143 

- Replaced `ignoreList` / `opList` / `friendList` / `bookmarkList` with string sets to make retrieval and removal more efficient
- Sets only contain lowercase names so all lookups can be made in a case-insensitive manner
- Removed the lists from the public interface, as they don't seem to be accessed from outside the class.

### DIFF
--- a/fchat/characters.ts
+++ b/fchat/characters.ts
@@ -33,20 +33,20 @@ class State implements Interfaces.State {
 
   friends: Character[] = [];
   bookmarks: Character[] = [];
-  ignoreList: string[] = [];
-  opList: string[] = [];
-  friendList: string[] = [];
-  bookmarkList: string[] = [];
+  ignoreSet: Set<string> = new Set<string>();
+  opSet: Set<string> = new Set<string>();
+  friendSet: Set<string> = new Set<string>();
+  bookmarkSet: Set<string> = new Set<string>();
 
   get(name: string): Character {
     const key = name.toLowerCase();
     let char = this.characters[key];
     if (char === undefined) {
       char = new Character(name);
-      char.isFriend = this.friendList.indexOf(name) !== -1;
-      char.isBookmarked = this.bookmarkList.indexOf(name) !== -1;
-      char.isChatOp = this.opList.indexOf(name) !== -1;
-      char.isIgnored = this.ignoreList.indexOf(key) !== -1;
+      char.isFriend = this.friendSet.has(key);
+      char.isBookmarked = this.bookmarkSet.has(key);
+      char.isChatOp = this.opSet.has(key);
+      char.isIgnored = this.ignoreSet.has(key);
       this.characters[key] = char;
     }
     return char;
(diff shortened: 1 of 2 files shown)


fchat/characters.ts
```diff
@@ -133,29 +136,26 @@ export default function (this: void, connection: Connection): Interfaces.State {
...
-      char.isChatOp = state.opList.indexOf(char.name) !== -1;
+      char.isIgnored = state.ignoreSet.has(key);
+      char.isChatOp = state.opSet.has(key);
     }
   });
   connection.onMessage('IGN', data => {
     switch (data.action) {
       case 'init':
-        state.ignoreList = data.characters.slice();
+        state.ignoreSet = new Set(data.characters);
```
FireUnderTheMountain, line 146: Ignore data can always be trusted to be lowercase?

fchat/characters.ts
```diff
@@ -106,24 +106,27 @@ export default function (this: void, connection: Connection): Interfaces.State {
...
       reconnectStatus = {
         status: state.ownCharacter.status,
         statusmsg: state.ownCharacter.statusText
       };
     for (const key in state.characters) {
       const character = state.characters[key]!;
-      character.isFriend = state.friendList.indexOf(character.name) !== -1;
-      character.isBookmarked =
-        state.bookmarkList.indexOf(character.name) !== -1;
+      character.isFriend = state.friendSet.has(key);
```
FireUnderTheMountain, line 128: I notice you don't use `toLowercase` on the values in `has` checks - You followed the code back and are certain these are always lowercase?

FireUnderTheMountain (commented):
I don't think the names needed to be changed; Variable names are for humans, and a list of your bookmarks is a list no matter whether its a Set or a Map or an array or a line in a file...

But the overall code quality is good.

fchat/characters.ts
little-voice: It is treated that way already in the existing code, so I didn't go out of my way to enforce it.

fchat/characters.ts
```diff
@@ -207,24 +207,24 @@ export default function (this: void, connection: Connection): Interfaces.State {
...
       case 'friendadd':
         if (character.isFriend) return;
-        state.friendList.push(data.name);
+        state.friendSet.add(data.name.toLowerCase());
         character.isFriend = true;
         if (character.status !== 'offline') state.friends.push(character);
         break;
       case 'friendremove':
-        state.friendList.splice(state.friendList.indexOf(data.name), 1);
+        state.friendSet.remove(data.name.toLowerCase());
```
FireUnderTheMountain, line 227: Arrays `remove`, Sets [`delete`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/delete). Strangely `pnpm watch` doesn't complain. Oversight?

little-voice: 🤦 good catch! Fixed.

little-voice:
Fair enough, I should have tested this more thoroughly. I think calling the contribution rotten and foul is a bit of an overreaction, but I get being annoyed.

PR #298, opened 2025-07-20
Update tab-switching shortcuts to use globalShortcut

### LINKED ISSUE #247: CTRL+Tab keyboard shortcut does not work in specific cases
**To Reproduce**

1. Open new tab with CTRL+T
2. Press CTRL+Tab (or CTRL+Shift+TAB)
3. Nothing happens
4. Switch tabs by clicking a tab
5. CTRL+Tab now works again

You can alternatively fix it by clicking inside the tab's web contents. So I assume the problem is simply because the focus is inside the wrong element (since the switch tab shortcut is somehow defined on either Index.vue or Chat.vue iirc)

**Screenshots**

- OS: Windows
- Version 1.32.1

little-voice:
Fixes #247


FatCatClient:
Have you looked into doing this with a local shortcut through a [hidden menu item](https://www.electronjs.org/docs/latest/api/menu-item#menuitemvisible) instead? 

I think that might be a less roundabout way of solving the problem, because blur/ focus events might trigger some really weird edge cases (especially on some Linux DEs), or act weird once multiple app windows come into play.

The alternative would be trying to see if it works if the keyboard event is handled someplace in the Window.vue component, but that'd only work if the input itself doesn't get eaten the moment you switch focus to the Index.vue root component inside the window.

little-voice:
I didn't try the hidden menu item option, I can give that a whirl though if we think that'd be better.  The blur/focus strategy was mentioned by an Electron maintainer [here](https://github.com/electron/electron/issues/27240#issuecomment-770861226), but I'm certainly willing to believe it does weird stuff in some cases/environments. I tried event handlers on the main window but couldn't find a way to make that work 🤷

FatCatClient:
If an Electron maintainer suggests it themselves, I'm more inclined to believe that it won't mess up sometimes. I'm mostly talking from experience where focusing and blurring went wrong while minimizing the window, but I also don't remember the exact circumstances that surrounded that bug.

@FireUnderTheMountain, do you remember the specifics better? I think you wound up fixing those issues, right?

FireUnderTheMountain:
After reading through little-voice's linked thread and a few adjacent Stack Overflow discussions, and then finding this snippet on the [API docs](https://www.electronjs.org/docs/latest/api/global-shortcut#globalshortcutregisteraccelerator-callback):

> When the accelerator is already taken by other applications, this call will silently fail. This behavior is intended by operating systems, since they don't want applications to fight for global shortcuts.

This implies inconsistency based on OS/desktop and running applications. I actually have no idea what the *practical* behavior is since trying to test this looks like a nightmare. But from that API snippet, if a user tabs out and registers Ctrl+Tab with chrome or anything else, they will lose Ctrl+Tab in Horizon 100% of the time, since it won't be registered.

So what happens if you globally bind Ctrl+Tab to another application and have a hidden menu item in Horizon bound to Ctrl+Tab? Does the global shortcut eat the key or propogate it, or is that OS/desktop/app dependent? In other words, is there a difference in behavior between this situation and the above?

---

Lastly, [this discussion](https://github.com/electron/electron/pull/45171) seems to suggest that electron can't even bind their own global shortcut on wayland yet, as they've only just barely implemented the modern portal paradigm and you have to go through a system config to actually set the bind (if I'm reading the post right).

So it seems like there's a ton of extra considerations for using electron's globalShortcuts that we might not have using an entirely contained method.

electron/browser_windows.ts
```diff
@@ -248,6 +248,24 @@ export function createMainWindow(
...
+  // Activate tab navigation through global shortcuts when main window is focused
+  window.on('focus', () => {
+    electron.globalShortcut.registerAll(
+      ['CmdOrCtrl+Tab', 'CmdOrCtrl+PageDown'],
+      () => window.webContents?.send('switch-tab')
+    );
+
+    electron.globalShortcut.registerAll(
+      ['CmdOrCtrl+Shift+Tab', 'CmdOrCtrl+PageUp'],
+      () => window.webContents?.send('previous-tab')
```
FireUnderTheMountain: Why `?`

electron/main.ts
```diff
@@ -474,6 +474,49 @@ function onReady(): void {
...
+                w.webContents.send('switch-tab');
+            }
+          },
+          {
+            accelerator: 'CmdOrCtrl+PageDown',
+            acceleratorWorksWhenHidden: true,
+            label: l('navigation.nextTab'),
+            click: (_m, w) => {
+              w instanceof electron.BrowserWindow &&
+                w.webContents.send('switch-tab');
```
FireUnderTheMountain, line 498: Why `instanceof electron.BrowserWindow`?

FireUnderTheMountain:
@little-voice  We're going to accept this one. Can you remove the unnecessary `acceleratorWorksWhenHidden` and then squash your commits down to something sensible and tidy?

little-voice:
Addressed change requests and squashed the commits.

FatCatClient:
For whatever reason, running the app in production mode on MacOS (and only on MacOS) causes these shortcuts to not work at all, though they work fine in developer mode. I don't know if this is something with Electron or the MacOS 26 beta, and whether or not that is a bug or feature on their end.

But that's not something you could have possibly known about, so I'll merge this PR anyway and fix that platform-specific bug myself. Thank you for your contribution.

PR #300, opened 2025-07-21
Set character card header z-index to be dimmed on modals

### LINKED ISSUE #282: Z-Index for tabs in the profile view is incorrect
**Describe the problem**
When you open a modal dialog on a profile, like the friends menu or the memo one, the persistent tab line appears above the dimmer, which makes it look like they're lighting up.

**Screenshots**

Image

_Before clicking modal._

Image

_After clicking modal._

- Version: 1.32.3-dev3 (but presumably ancient)
- Theme: All

**Additional context**
This should be an easy fix. Just double check it with [the Z-indexes that Bootstrap 4](https://getbootstrap.com/docs/4.1/layout/overview/#z-index) defines for things like modals, etc. I _think_ the tiny modals on the character page are stock Bootstrap ones as well, but I could be wrong.

DerEchteDimenzio:
Fixes #282

Only change to header behaviour is that hover popover of kinks now appear above the header, instead of below. Due to popver z-index being 1060 and modal ones being 1040, hence why the header's z-index must be lower. 

Before:
Screenshot 2025-07-21 205421

After:
Screenshot 2025-07-21 205435

### DIFF
--- a/site/character_page/character_page.vue
+++ b/site/character_page/character_page.vue
@@ -1047,7 +1047,7 @@
   .character-card-header {
     position: sticky;
     top: -1rem;
-    z-index: 10000;
+    z-index: 1000;
     background: var(--headerBackgroundMaskColor) !important;
   }
 


FatCatClient:
Accidentally merged this into main instead of dev, whoops. Don't mind me having reverted that and merging it w/ 2c82aaa8ad4e6e62606e679f6338689ba87558b8 instead :^)

Issue #229, opened 2025-06-19
Expirable statuses


FatCatClient:
**Describe the solution you'd like**
When setting a status, we can have another select dropdown where you can enter how long your status should stay up for-- with the default behaviour being no expiration like we have now.

![Image](https://github.com/user-attachments/assets/21974965-36d2-457c-a821-82f2c139d906)

PR #311, opened 2025-07-26
Add status timeout feature

DerEchteDimenzio:
Implements #229

Feedback appreciated, especially regarding interface and file structure.


FatCatClient:
How does this interact with the idle timer? Does setting a status for X minutes reset your idle status if the timer for the latter is X-1 minutes?

I'll look a bit into the actual implementation later though, doubly so since you wanted some feedback on that.

Issue #310, opened 2025-07-26
Shortcut for BBCode preview


FatCatClient:
**Is your feature request related to a problem? Please describe.**

Can't be bothered to write an extensive treaty on this, because it's a simple idea.

<img width="1158" height="176" alt="Image" src="https://github.com/user-attachments/assets/9773ffff-487c-4ec7-b389-e775b7c628c2" />

This button needs a shortcut (preferably CTRL+Shift+P), the same way the other buttons in that list do.

PR #312, opened 2025-07-27
Add preview keyboard shortcut

DerEchteDimenzio:
Implements #310

### DIFF
--- a/bbcode/Editor.vue
+++ b/bbcode/Editor.vue
@@ -69,7 +69,7 @@
           @click="previewBBCode"
           class="btn btn-light btn-sm"
           :class="preview ? 'active' : ''"
-          :title="preview ? 'Close Preview' : 'Preview'"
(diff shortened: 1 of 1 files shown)


FatCatClient:
Would you mind doing this through the list of buttons in the ``bbcode/editor.ts`` file instead of as a separate event handler? You'd have to do a bit of a refactor in both that file and the Vue component's event handler to allow for buttons that don't have a tag defined for them (and can have the shift modifier applied), but it would solve the following problems:
- An extra window-level event handler adds a bit of extra performance overhead, but it also controls every instance of the bbcode editor at once:

https://github.com/user-attachments/assets/6bee2623-6de8-4a93-b529-f08fcbf4c5e3

- Differences between shortcuts for Mac users (who'd be using Cmd+Shift+P instead of the Ctrl key), both for the key event itself, and for the tooltip
- Would be cleaner code-wise, since all of the editor's shortcuts are defined in the same spot.

Don't worry about this moving the help button to the end of the list, this might straight up be preferable.

PR #384, opened 2025-09-03
Fix archive export corruption

### LINKED ISSUE #213: Unusual characters (emoji, styled text) corrupt exported log ZIP files
**Describe the bug**
When the logs of a character contain emoji or unusual characters like styled text, exporting the logs as a ZIP file could result in a corrupted ZIP file. Worst case: It can't be used at all. Best case: It can be repaired and only some logs will be missing from the repaired ZIP file.

**To Reproduce**

1. Create a private room
2. Post text in that channel like this: Ⲥ𝘏𐌉𝖭𝝤𝙍𝗦ꓴ𝝪
3. Open the log viewer
4. Export logs for the character
5. Try to open the resulting zip file and observe errors

**Expected behavior**
Characters that that aren't allowed as folder or file names in ZIP files should be filtered or replaced during export.

- OS: Windows or Linux
- Version: 1.30.3 (this bug has existed for a long time and is present in - AFAIK - all public releases of F-Chat)

**Additional context**
I have reported this on the official site but there is no way they're going to fix it on their end.
I'm not going to make any promises, but I might attempt fixing it myself. We'll see.

greyhoof:
fix: Replace custom zip solution with adm-zip package to fix export corruption when logs contain certain characters

This would resolve #213 

This PR is a bit of a "suggestion."

I have added the adm-zip package, which is a small package that utilizes Node's zlib to compress archives. I've chosen this because it has no other dependencies and it supports creating archives in memory, as well as streaming the result to a buffer, which allows triggering the download like Horizon did before.

**Upsides**:
- Archives no longer get corrupted when logs and/or channel names contain certain characters
- Emoji (at least some) are correctly represented in folder names within the zip file

**Downsides**:
- adm-zip does not support selecting a compression level, and it always applies compression to some degree, which increases processing time

If you don't like this solution, feel free to reject and I'll try something else. ✌️

### DIFF
--- a/chat/Logs.vue
+++ b/chat/Logs.vue
@@ -318,6 +318,49 @@
       });
     }
 
+    sanitizeConversationName(name: string): string {
+      /*
+       * Replace characters that are forbidden in paths with an underscore.
+       * This list should cover Unix, Windows and macOS.
+       * Files and folders also may not end with spaces.
+       */
+      let sanitizedName = name.replace(/[\/<>:"\\|?*.]/g, '_').trimRight();
+
+      /*
+       * For Windows, certain names are forbidden, too.
+       */
+      if (
+        [
+          'CON',
+          'PRN',
+          'AUX',
+          'NUL',
+          'COM1',
+          'COM2',
+          'COM3',
+          'COM4',
+          'COM5',
+          'COM6',
+          'COM7',
+          'COM8',
+          'COM9',
+          'LPT1',
+          'LPT2',
+          'LPT3',
+          'LPT4',
+          'LPT5',
+          'LPT6',
+          'LPT7',
+          'LPT8',
+          'LPT9'
+        ].includes(sanitizedName)
+      ) {
+        sanitizedName += '_';
+      }
+
+      return sanitizedName;
+    }
+
     downloadDay(): void {
       if (
         this.selectedConversation === undefined ||
@@ -326,7 +369,7 @@
       )
         return;
       const html = Dialog.confirmDialog(l('logs.html'));
-      const name = `${this.selectedConversation.name}-${formatDate(new Date(this.selectedDate))}.${html ? 'html' : 'txt'}`;
+      const name = `${this.sanitizeConversationName(this.selectedConversation.name)}-${formatDate(new Date(this.selectedDate))}.${html ? 'html' : 'txt'}`;
       this.download(
         name,
         `data:${encodeURIComponent(name)},${encodeURIComponent(getLogs(this.messages, html))}`
@@ -349,7 +392,7 @@
         );
       }
       this.download(
-        `${this.selectedConversation.name}.zip`,
+        `${this.sanitizeConversationName(this.selectedConversation.name)}.zip`,
         URL.createObjectURL(new Blob([zip.toBuffer()]))
       );
     }
@@ -374,7 +417,7 @@
             date
           );
           zip.addFile(
-            `${conv.name}/${formatDate(date)}.${html ? 'html' : 'txt'}`,
+            `${this.sanitizeConversationName(conv.name)}/${formatDate(date)}.${html ? 'html' : 'txt'}`,
             Buffer.from(getLogs(messages, html), 'utf-8')
           );
         }
--- a/chat/Logs.vue
+++ b/chat/Logs.vue
@@ -405,11 +405,17 @@
         return;
       const zip = new AdmZip();
       const html = Dialog.confirmDialog(l('logs.html'));
+      const existingConversationNames = new Array<string>();
       for (const conv of this.conversations) {
         const dates = await core.logs.getLogDates(
           this.selectedCharacter,
           conv.key
         );
+        let sanitizedConvName = this.sanitizeConversationName(conv.name);
+        while (existingConversationNames.includes(sanitizedConvName)) {
+          sanitizedConvName += '_';
+        }
+        existingConversationNames.push(sanitizedConvName);
         for (const date of dates) {
           const messages = await core.logs.getLogs(
             this.selectedCharacter,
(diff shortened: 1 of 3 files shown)


greyhoof:
Hold on, I was missing something with this. I'll have to fix it first.

FatCatClient:
I've been meaning to test this just now, but it turns out that **something** completely broke exporting all your logs for a specific character at once in the dev branch. 

So, we'll get on that once that issue is fixed.

FireUnderTheMountain:
Strange question, have you tested results from a version where you simply change `if (c > 0xD800 && c < 0xD8FF)` to `if (c >= 0xD800 && c <= 0xDBFF)` in Zip.fileAdd?

FireUnderTheMountain:
Anyways, this is what tipped me off:
```ts
Buffer.from(getLogs(messages, html), 'utf-8')
```
You're not using "a better zipper", you're also using a parsed input first... Why can't do we do the same with the built-in zip?

The question that haunts me is, why does the zipper work fine for *most* files but not a couple with special high unicode code points?

I used [this converter](https://r12a.github.io/app-conversion) to make a character with utf-16 code `D900 DD21` which is outside the [mistyped] range and it's incorrectly read as 6 bytes by the `getByteLength` function that show you how much you've typed in chat. That same function is used in the zipper to generate an offset from a file's name, which means that character breaks log zipping when it's in a channel name. This is the first time I've actually been able to make a corrupt zip! And on linux, at that. :)

This is, of course, fixed by replacing `getByteLength` with `new TextEncoder().encode().length` (TextEncoder being the browser built-in to replace node's Buffer.from for getting utf-8 bytes). It was also fixed by using the correct high surrogate pairs range in getByteLength. There's still garbage in the channel filename, but it's consistent. It also makes the chat-typing counter count correctly (4 bytes intead of 6).

I'm still not absolutely certain, but I believe fixing the byte range to cover all of the high surrogate pairs (the 0xD800 <= c <= 0xDBFF range) in **both** implementations would resolve the problem. If it doesn't, do you happen to know of specific emojis or characters that cause breakage? They can be put into that site to get their utf-16 code so we can see if there's a pattern.

Sorry if you read all that, I'm going to sleep now 🥳

FireUnderTheMountain:
`𝘏𝖭𝝤𝙍𝗦𝝪`: Doesn't seem to cause any issues.

`Ⲥ`: Doesn't cause me any issues in a chat text message.

`ꓴ`:  Doesn't seem to cause any issues.

`𐌉`: Starting bytes are 0xD800 - it causes corruption reliably and is fixed by changing the lower bound of the range to include the missing lowest value.

--
> line 26 in zip.ts and line 24 in common.ts

Correct. Set my tree to development, refreshed dependencies, and began testing.

Here was the only fix I needed to make to get working zips with those specific characters in channel name and chat message in a full-character export:
```diff
diff --git a/chat/common.ts b/chat/common.ts
index 9b4ac026..0cab66b7 100644
--- a/chat/common.ts
+++ b/chat/common.ts
@@ -21,7 +21,7 @@ export function getByteLength(this: any | never, str: string): number {
   let byteLen = 0;
   for (let i = 0; i < str.length; i++) {
     let c = str.charCodeAt(i);
-    if (c > 0xd800 && c < 0xd8ff)
+    if (c >= 0xd800 && c < 0xd8ff)
       //surrogate pairs
       c = (c - 0xd800) * 0x400 + str.charCodeAt(++i) - 0xdc00 + 0x10000;
     byteLen +=
diff --git a/chat/zip.ts b/chat/zip.ts
index b1c37efe..0c1e3ca2 100644
--- a/chat/zip.ts
+++ b/chat/zip.ts
@@ -23,7 +23,7 @@ export default class Zip {
     const nameLength = getByteLength(name);
     for (let i = 0, strlen = content.length; i < strlen; ++i) {
       let c = content.charCodeAt(i);
-      if (c > 0xd800 && c < 0xd8ff)
+      if (c >= 0xd800 && c < 0xd8ff)
         //surrogate pairs
         c = (c - 0xd800) * 0x400 + content.charCodeAt(++i) - 0xdc00 + 0x10000;
       let l =
```

Didn't even fix the upper bound missing the majority of starter surrogates yet.

**However** I only tested on linux; linux doesn't throw a fit when folder names contain wacky characters (only the folder separator and the null byte can't be part of a filename).

Therefore, I believe this may still be an issue on windows if they're going to return invalid windows folder names. If that phrase, with that fix, still produces corrupt zips on windows, then I'm with you - this is stupid, the easy solution is nice, lets just go with it. Easy solutions to "windows is being dumb again" are *always* appreciated.

greyhoof:
Yeah, I still meant to make some changes regarding path separators and such. I'm not 100% on which characters are relevant here, since they only play a role in conversation names, I think.

Forward slash is a given, backslash shouldn't matter (even on Windows) IIRC. Null byte is probably not usable in a channel name, lol. Is there anything else?

FatCatClient:
> Does the following sound good to you?
> 
> * Individual characters -> replace with underscore
> 
> * Magic names -> append underscore (Windows allows that)

Sounds good to me.

greyhoof:
Does this do the trick for y'all? :)
I tested it with a few path characters in channel names and it worked fine.

An edge case I didn't cover was channel names that happen to have forbidden characters in the same places but are otherwise called the same. They would end up with the same sanitized name so there would be a mixup with logs.
So "room \ with / path * characters" and "room * with ? path . characters" would both become "room _ with _ path _ characters"

Do you consider this situation to be rare enough that we can ignore it or should I implement a safety precaution against that?

FatCatClient:
> Do you consider this situation to be rare enough that we can ignore it or should I implement a safety precaution against that?

I mean I do– because when can you actually expect someone to run into something like this, but if it's not too much trouble we might as well go for the full clean implementation.

FatCatClient:
Sounds good to me. It's a super niche case, but we might as well cover those bases too.

@FireUnderTheMountain did you still have some thoughts on the UTF-8/ UTF-16 shenanigans we talked about?

PR #388, opened 2025-09-05
Adjustments to modern chat UI - Probably needs optomising and definitely needs toggles.

BootsieWootsie:
So, putting this here while I work on it a little more but this fixes /me interactions being a little janky. I've put a few screenies in the discord earlier for that.

The primary thing this includes is grouping (needs toggle implemented as of time of writing) like discord does, it's set to group every message someone makes 2 minutes after their last AS LONG AS nothing else interrupts between it. Needs some more testing and all. 
The primary issue with this is that it stores the entirety of the previous message in the current messageView every time. I don't imagine the RAM hit will be significant but it may be worth figuring out something better. Either me or someone who is signficantly smarter than me.

I'd work on it more before sending but it is 1 AM and this is for something that's likely going in 1.34 anyway so there should be time to fix it up. Can gather some more screens and put them in here later too.

### DIFF
--- a/chat/ConversationView.vue
+++ b/chat/ConversationView.vue
@@ -300,12 +300,13 @@
       @scroll="onMessagesScroll"
       style="flex: 1; overflow: auto; margin-top: 2px"
     >
-      <template v-for="message in messages">
+      <template v-for="(message, i) in messages">
         <message-view
           :message="message"
           :channel="isChannel(conversation) ? conversation.channel : undefined"
           :key="message.id"
           :classes="message == conversation.lastRead ? 'last-read' : ''"
+          :previous="messages[i - 1]"
         >
         </message-view>
         <span
--- a/chat/Logs.vue
+++ b/chat/Logs.vue
@@ -117,10 +117,11 @@
       @scroll="onMessagesScroll"
     >
       <message-view
-        v-for="message in displayedMessages"
+        v-for="(message, i) in displayedMessages"
         :message="message"
         :key="message.id"
         :logs="true"
+        :previous="messages[i - 1]"
       ></message-view>
     </div>
     <div class="input-group" style="flex-shrink: 0">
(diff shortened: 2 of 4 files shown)


FatCatClient:
> The primary issue with this is that it stores the entirety of the previous message in the current messageView every time. I don't imagine the RAM hit will be significant but it may be worth figuring out something better. Either me or someone who is signficantly smarter than me.

If you want to compare that, you should be able to just use the dev tools profiler for that. It's under the 'Memory' tab. But you're right, I haven't looked at the actual changes yet, but based on the files touched those aren't likely going to be terribly significant changes in memory usage. THat's more reserved for the kind of thing that'd touch every message kept in memory, rather than just whatever group of messages happens to be on screen in the ConversationView component.

> I'd work on it more before sending but it is 1 AM and this is for something that's likely going in 1.34 anyway so there should be time to fix it up. Can gather some more screens and put them in here later too.

Yeah exactly. There's no rush, because it's not like that one's right around the corner.

BootsieWootsie:
I have already noticed another bug for tomorrow, dates in logs sometimes pull the message into them.

BootsieWootsie:
The bug I mentioned last night is still present, there's another one that happens occasionally when swapping channels after swapping settings for modern/classic and locks you out of the current tab of the client until you relog. 

Far as I can tell, the second one relates to cleanup functions being a little buggy with modern, still tracing the first but I have a simple css fix (just make bbcode not display in time element) for it in worst case. Wouldn't be ideal to fix it with CSS since it doesn't address the root of the issue though.

BootsieWootsie:
Both of those _should_ be fixed now, gonna keep testing to be sure on the lockouts though since that one was pretty bad! I did notice a bug with logs not recognising the l function (though it doesn't appear to have caused any ill effects) despite having imported it but I suspect that isn't the fault of this since I didn't touch that.

PR #405, opened 2025-09-11
More Modern chat UI adjustments

BootsieWootsie:
As said on the tin, this is primarily a bunch of changes to Modern Chat UI plus that one extra change.

- Removes own message background (opinionated change, would love to hear back on it before pushing this into dev proper)
- Adds time on hover for modern
- Changes spacing, sizing, padding again.
- Changes how borders work a bit, messages should no longer move around when you tab back in and clear unread and should stay consistent in general.
- Adds a border to the left side of chat messages that is invisible, particularly there for the new border on the left of highlighted messages (probably needs adjustment to be brighter still)
- As mentioned, Classic Chat UI avatars no longer use Horizon Portraits (can split this to a seperate PR if wanted)
- Fixes(?) a bug in common where noDate on formatTime doesn't do as advertised. I'm not sure as of writing if that's used anywhere with that variable set to true other than where I've used it but will be double checking to be sure. If it is, I'll just split it to it's own function for short time.


chat/UserView.vue
```diff
@@ -244,6 +244,9 @@
     @Prop({ default: false })
     readonly avatar: boolean = false;
 
+    @Prop({ default: false })
+    readonly avatarOrig: boolean = false;
```
FatCatClient, line 248: For an outward facing parameter, I don't think this is a good name. Something like ``useOriginalAvatar`` gets the point across easier, especially since our current setup won't allow for TSDoc-style documentation to properly work in Vue SFCs.

chat/common.ts
```diff
@@ -7,10 +7,14 @@ export function profileLink(this: any | never, character: string): string {
   return `https://www.f-list.net/c/${character}`;
 }
 
-export function characterImage(this: any | never, character: string): string {
+export function characterImage(
+  this: any | never,
+  character: string,
+  original: boolean = false
```
FatCatClient, line 13: Same as the parameter in UserView.vue. TSDoc annotations might be a nice addition regardless, but that's obviously not your fault.

scss/_chat.scss
```diff
@@ -470,10 +491,31 @@
...
     }
+    .message-own {
+      background: transparent;
+    }
     &.message-own .message-content .bbcode {
       background: transparent;
     }
   }
+  .message-highlight {
+    border-left: solid 2px
```
FatCatClient, line 502: Not a big fan of setting rules with !important tags compared to increasing the specificity levels. That tends to get really annoying when exceptions to those rules are created.

chat/common.ts
BootsieWootsie: True, I'll likely forward both those changes when pulling it to it's own PR there so it's a little more descriptive

BootsieWootsie:
> I'll add real quick, that the last read message bottom border causing things to jump around is only really a problem because (for some reason), focusing into the web page causes that specific conversation to be considered "read". The other two moments this happens is when you intentionally interact (mark all as read button, or sending a message), for which I don't personally consider that a real problem.
> 
> Now that I think about it, it might make way more sense to just have it not set the conversation to read once you focus back into a tab. Discord doesn't do that either, and I don't think anyone complains there.

I'm slightly wondering if we should do this specifically with modern or set it up for both. I believe that it's handled in ConversationView.vue and conversation.ts for when it should and should not have it. Might take a little bit of digging to get it going proper. Though, yeah, I think requiring actual interaction to change it is probably ideal.

> Also now that you're there, maybe it's worth looking at how console messages are rendered in the actual console, inside a conversation and in the log view. There's definitely inconsistencies there. (Ie the console tab always shows in classic view like it should, but in the other two contexts the messages might not look correct anymore).

I did attempt to make them look the same but that was before the changes to the time so they look a little buggered now. Deffo need to take a look at that.

FatCatClient:
> I'm slightly wondering if we should do this specifically with modern or set it up for both. I believe that it's handled in ConversationView.vue and conversation.ts for when it should and should not have it. Might take a little bit of digging to get it going proper. Though, yeah, I think requiring actual interaction to change it is probably ideal.

Just do it for both, I'd say. iirc the instance in ConversationView.vue is the one that triggers on the blur event, which is the only one I'd get rid of. It fires off even when you do things like clicking the buttons in the tab window bar. The other instances of the last read property changing happen when doing things like switching away to a different conversation, sending a message or clicking the mark as read button. All of those are fine and won't cause any jarring jumps.

In fact, I'd personally probably prefer that was the only change and the actual message border CSS can be left alone. It's probably fine that there is a bit of spacing between the unread indicator and messages.

PR #411, opened 2025-09-16
Original avatars in classic chat view

BootsieWootsie:
Moved from https://github.com/Fchat-Horizon/Horizon/pull/405/
This just makes the avatar at the side of UserView use the original avatar instead of the Horizon portrait when you are in classic view.


BootsieWootsie:
I'm stupid and nearly sent to main

PR #415, opened 2025-09-20
Reminders on which awful piece of the cache system does what.

FireUnderTheMountain:
Better late than never.

One of the comments is in there twice because it needed to be in order to show up properly in vscodium.

There's an addendum on tsdoc comments that these notes were imported, that way when you stumble across them in the future or as part of an exported doc you won't be wondering why they have mild inaccuracies or bad opinions. Remove those import notices as you rewrite/invalidate the markup.

### DIFF
--- a/interfaces.ts
+++ b/interfaces.ts
@@ -1,3 +1,10 @@
+/**
+ * The most simple character structure, used primarily for listing your characters on the log-in selector.
+ *
+ * Character name, unique ID, and whether or not the character exists.
+ *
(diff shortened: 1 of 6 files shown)


interfaces.ts
```diff
@@ -51,6 +58,13 @@ export interface Infotag {
   infotag_group: number;
 }
 
+/**
+ * Public character page. Why does this extend SimpleCharacter? Investigate usage.
```
FatCatClient, line 62: I'm going to assume this extends SimpleCharacter because of something that was designed for the old 2.0 site.

Issue #266, opened 2025-07-08
Exporting all logs to HTML freezes if you have too many logs


FatCatClient:
**Describe the bug**
I have no idea where the cutoff for it is, and I've only seen it happen on characters with over 17 MB worth of log files

** Repro **

1. Go to log in screen
2. Click logs and select a character
3. Click the download button for your logs, and then select HTML logs
4. Depending on how severe it gets, it can even crash the render process

**Version and system info (please complete the following information):**

- OS: Windows
- Version 1.32.2

greyhoof:
While working on the zip issue, I noticed this, too.

I have a character with an absurd amount of logs and during the export process (while Horizon grabs the eicons from the server) the client will stop the export before it crashes due to an out-of-memory event.

I'm spitballing here, but it might just come down to a combination of the size of the logs itself as well as including the eicons - and how Horizon performs the export & downloads the exported file.

<img width="554" height="209" alt="Image" src="https://github.com/user-attachments/assets/c885ffac-bda3-4127-93cb-844524837956" />
<img width="319" height="20" alt="Image" src="https://github.com/user-attachments/assets/249ecc04-2089-4d0a-9b17-6fd62fdd7397" />

Maybe a reasonable solution would be to track the amount of processed logs and split it up every x amount of files, triggering a download each time?

FatCatClient:
> Maybe a reasonable solution would be to track the amount of processed logs and split it up every x amount of files, triggering a download each time?

It'd definitely reduce some of the convenience of being able to save everything in one .zip file, so if I were to look at this I'd first try and see if there's something about the current process that can be smoothened out. Does it, for instance, keep the whole conversation/ log object in memory for every conversation? Stuff like that. Could we work with temp, split files, so we don't have to keep everything in costly JS memory, and then merge them together at the end? 

I haven't really had a lot of reasons to properly look into the zipping code yet, so I'm spitballing just as much as you are But as far as I know, it hasn't seen any meaningful changes since its first iteration and maybe these ideas are possible.

But if that turns out to be a dead end, then I think that's a compromise worth making. I'd rather that for these extreme cases it doesn't work in an ideal way, than have it not work at all. And I think that few people would disagree with that.

greyhoof:
> Does it, for instance, keep the whole conversation/ log object in memory for every conversation?

As far as I'm aware, It does. The way the zip is created, it performs that action entirely in-memory.

[...]

PR #434, opened 2025-10-02
Fix cleanup not being called when exporting HTML logs

greyhoof:
fix: Ensure that the log export cleans up elements after fetching the HTML generated from logs

This reduces RAM usage when exporting logs. It should also _technically_ fix the log export freezing/crashing issue #266
But I still think that needs a more involved approach to be considered properly fixed.

### DIFF
--- a/chat/Logs.vue
+++ b/chat/Logs.vue
@@ -175,7 +175,12 @@
                 }
               : undefined,
             html
-              ? t => `${core.bbCodeParser.parseEverything(t).innerHTML}`
+              ? t => {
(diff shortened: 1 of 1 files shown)


CodingWithAnxiety:
I misunderstood this PR, I apologize.

I think this is good to merge, but I'll wait for the other dev's thoughts.

FatCatClient:
It took me 8 minutes to export one character's logs into HTML with the old version, while testing this PR did it in 2 minutes.

I think that says enough.

PR #439, opened 2025-10-05
fix: add invidious to client

### LINKED ISSUE #419: Adblocking doesn't work (properly) for Youtube urls
_This is a mirror of https://github.com/Frolic-chat/Frolic/issues/38, so that we can track this here too– and keep it on the list for things to fix before we can drop 1.34.0._

**Version**: [b7278a7bc0525ba077f91a0074802c559c35284d](https://github.com/Frolic-chat/Frolic/commit/b7278a7bc0525ba077f91a0074802c559c35284d) built from source
**Platform**: Ubuntu

_Occurs in Horizon versions starting with 1.34.0-dev1 too_

Realized this while trying to get https://github.com/Fchat-Horizon/Horizon/pull/417 working, but apparently Youtube links just don't filter ads in the previewer, or at least not consistently. Though for that PR it's running on Ghostery 2.11.0^, but for the 2.8 versions it's pretty much the same story.

Wish I could have a more extensive bug report than this, but I haven't figured out what the exact problem is either. Maybe we need to reference different block lists? I tried it with [this](https://www.youtube.com/watch?v=mDpvHOsXC8g) video specifically though.

CodingWithAnxiety:
Readds an invidious instance to the client, allowing us to avoid ads.

Closes #419


CodingWithAnxiety:
oh i'm fucking stupid.

PR #443, opened 2025-10-07
feat: add logging settings to conversation settings

### LINKED ISSUE #385: Expanded Logging Options
Copy-paste from Discord:

> Option to disable logging on channels, to: reduce disk space, security i guess, less IO operations helping reduce log sync choking when log is set to OneDrive (locks files during upload -- f-chat can error out if it can't write to it, possibly for set retries etc. but who knows, so if channel is spammy it can crash the tab)

> Option to enable or disable logging per room is probably best way to implement this, enabling mods to still log channels for moderation purposes.

CodingWithAnxiety:
Adds per-channel logging options, allowing you to disable / enable it on a per-channel basis.

Closes #385.

### DIFF
--- a/chat/ConversationSettings.vue
+++ b/chat/ConversationSettings.vue
@@ -101,6 +101,26 @@
         </option>
       </select>
     </div>
+    <div class="mb-3">
+      <label class="control-label" :for="'enableLogging' + conversation.key">{{
+        l('conversationSettings.enableLogging')
+      }}</label>
+      <select
+        class="form-select"
+        :id="'enableLogging' + conversation.key"
+        v-model="enableLogging"
+      >
+        <option :value="setting.Default">
+          {{ l('conversationSettings.default') }}
+        </option>
+        <option :value="setting.True">
+          {{ l('conversationSettings.true') }}
+        </option>
+        <option :value="setting.False">
+          {{ l('conversationSettings.false') }}
+        </option>
+      </select>
+    </div>
   </modal>
 </template>
 
@@ -125,6 +145,7 @@
     horizonHighlightUsers!: string;
     joinMessages!: Conversation.Setting;
     defaultHighlights!: boolean;
+    enableLogging!: Conversation.Setting;
 
     load(): void {
       const settings = this.conversation.settings;
@@ -134,6 +155,7 @@
       this.joinMessages = settings.joinMessages;
       this.defaultHighlights = settings.defaultHighlights;
       this.horizonHighlightUsers = settings.horizonHighlightUsers.join(',');
+      this.enableLogging = settings.enableLogging;
     }
 
     submit(): void {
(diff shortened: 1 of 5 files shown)


CodingWithAnxiety:
kms next time prettier fails genuinely

chat/conversations.ts
```diff
@@ -33,6 +33,28 @@ import isChannel = Interfaces.isChannel;
  */
 const CONVERSATION_CACHE_UPDATE_FREQ_IN_MS = 1000;
 
+
+function shouldEnableLogging(
```
FatCatClient, line 37: Considering that every method call to this is just this bit in the middle of the message parsing
```ts
if shouldEnableLogging() {
         core.logs.logMessage();
    }

```

I would have just made this a member function of the abstract conversation class that also calls the logMessage() method after passing the checks, with the logic to check if the message itself is an ad worked in as well.

Could just be something like 
```ts
function logMessage(message: Message) {
    if(settings.shouldLog) {
    // etc.

    core.logs.logMessage(...)
    }
}
```

chat/conversations.ts
```diff
@@ -33,6 +33,28 @@ import isChannel = Interfaces.isChannel;
...
+  // If explicitly set to True or False, use that
+  if (loggingSetting === Interfaces.Setting.True) {
+    return true;
+  }
+  if (loggingSetting === Interfaces.Setting.False) {
+    return false;
+  }
+  
+  // Otherwise (Default), use global settings
+  if (isAd) {
```
FatCatClient, line 52: Will this check ever be hit if you have the setting set to a non-default value? Seems to me like it will now only check for the ad setting if you have the conversation-level setting set to default, which I doubt is intended.

I'll add that I haven't tested that specific test case myself, but did you?

chat/ConversationSettings.vue
```diff
@@ -101,6 +101,26 @@
...
+    <div class="mb-3">
+      <label class="control-label" :for="'enableLogging' + conversation.key">{{
+        l('conversationSettings.enableLogging')
+      }}</label>
+      <select
+        class="form-select"
+        :id="'enableLogging' + conversation.key"
+        v-model="enableLogging"
+      >
+        <option :value="setting.Default">
```
FatCatClient, line 113: This isn't a comment about your work, but I just want to say that when we do global settings that this shit should really just be a generic component instead of something we copy paste a bajillion times.

chat/ConversationSettings.vue
```diff
@@ -151,7 +173,8 @@
 
         joinMessages: this.joinMessages,
         defaultHighlights: this.defaultHighlights,
-        adSettings: this.conversation.settings.adSettings
+        adSettings: this.conversation.settings.adSettings,
+        enableLogging: this.enableLogging
```
FatCatClient, line 177: Maybe it's better for consistency if this has the same name as the character-level setting. This is the same with other settings that can be overridden per conversation like ``joinMessages``, ``highlight``, ``highlightWords``, etc.

PR #460, opened 2025-10-16
Catch og:image and twitter:image before grabbing preview image.

snowsune:
Adds just a few lines, checking for og:image and twitter:image (and more if wanted!) before loading the first image. Fixes previews broken on sites that don't prepend some image to show first!


snowsune:
Oh! yes sorry, didn't see this at first. Here you go: https://horizontest.snowsune.net/

PR #512, opened 2025-11-14
chore: Upgrade Electron to v38.7.0

FatCatClient:
We originally had this planned for 1.34.0, but then an upstream bug caused issues with notifications on Windows.

Let's do it for realises this time.


CodingWithAnxiety (approved):
Looks good. Approving.

PR #516, opened 2025-11-16
Adds 'All Friends' tab in the console's user list

FatCatClient:
~This was spun off from the [character friends branch](https://github.com/Fchat-Horizon/Horizon/tree/feat/per-character-friends), so it's probably smart to merge that into development before looking at this.~

~But this is a separate feature, so it shouldn't be thrown into the same PR.~

This also adds a property to UserView that lets you determine whether or not this particular instance should bother fetching a name color.

Issues to resolve:

- [x] Friends list can show duplicate entries, if you are friends with said character on multiple of your characters
- [x] Having the friends list respect the Character Friends settings would be nice

### DIFF
--- a/chat/UserList.vue
+++ b/chat/UserList.vue
@@ -42,7 +42,7 @@
       <h4 v-if="friends.length > 0">
         {{
           l(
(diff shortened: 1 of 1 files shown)


FatCatClient:
Or we can just switch the target branch back and forth :^)

PR #542, opened 2025-12-13
Implements infinite loading + scrolling for e-icon search results

freenutsxd:
Allows you to scroll through **all** e-icon search results instead of just a flat 301 maximum. For performance purposes it only loads 77 results at a time (1.5 pages) until you scroll to the bottom, which should help speed up loading while we're at it.

Technically the random and blank search results only scroll up to 2000 results, but I _really_ don't think anyone's scrolling further than that

### DIFF
--- a/bbcode/EIconSelector.vue
+++ b/bbcode/EIconSelector.vue
@@ -736,6 +736,14 @@
       this.runSearch();
 
       this.refreshing = false;
+
+      this.$nextTick(() => {
+        // reattach scroll listener after refresh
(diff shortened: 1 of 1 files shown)


freenutsxd:
Done and done, I'm kinda surprised I didn't check that refreshing case (I've never used the refresh button in normal usage). The issue was because the results container was unloaded every time the eicons were refreshed, breaking the scrolling, but now it's good. Couldn't find any other case where the scrolling would break, either

Issue #543, opened 2025-12-14
Leftmost icon pixel gets cropped in PMs


min-xy:
The PMs window will cut a single pixel off the edge of some icons. To fix, a left-padding can be edited to a 1 from 0.

Current appearance:
<img width="609" height="385" alt="image" src="https://github.com/user-attachments/assets/f1e5d9ae-93b3-4f03-ac45-574ad52e3f70" />

How it should look:
<img width="624" height="427" alt="image" src="https://github.com/user-attachments/assets/c66feb21-dc35-4217-a802-879b880d960c" />

Windows 11. 
Version 1.35.4

I believe this should just be in the CSS in ChatView.vue:
.list-group.conversation-nav {
...
&.item-private {
padding-left: 0;

And flipping that 0 to a 1.

I got the screenshots by opening the dev tools and editing:
.list-group.conversation-nav .list-group-item.item-private

PR #544, opened 2025-12-16
Prevent icon cutoff

min-xy:
Icons in private messages that are 100x100 have their leftmost pixel cut off.

Added a single pixel of left padding. While testing, noticed that the grey bubble could also be smaller than the chat icon at some font sizes, so added top and bottom padding too.

Examples of old behavior:
image
image

After my changes:
image
image

I know it probably seems trivial and pointless, but it is annoying to see an icons outline only outlining 3/4ths of the icon! I tested the behavior with multiple font sizes, and I didn't see any change to how it was at 24, but at 14 was where that first case of the icon being larger than the box it lived in was, and the left icon issue was most common at that size.

### DIFF
--- a/chat/ChatView.vue
+++ b/chat/ChatView.vue
@@ -937,9 +937,9 @@
         }
       }
       &.item-private {
-        padding-left: 0;
-        padding-top: 0;
-        padding-bottom: 0;
+        padding-left: 1px;
+        padding-top: 1px;
+        padding-bottom: 1px;
 
         .online-status {
           padding-left: 1px;


FatCatClient:
I'm sure this is fine and all, but could you do me a favour and send a better screenshot than that? It's a bit hard for me to see the difference here 😅

min-xy:
Sure!

I really zoomed in to do this. This is the before, size 14 font. The border is the same all the way around. As you can see, we lose one pixel (that is at ALL) display sizes.

The icon hanging over the grey box is mostly at 14. But the behavior doesn't change after adding the padding, it just removes that case.

Here's the before:
image

And the after (also at 14):
image

I know it seems super minor but my eyes always jump to the missing borders and it feels wrong. And I just submitted this to close out #543 which I had logged. 😅

PR #553, opened 2025-12-18
Convert character_select.vue to Composition API

### LINKED ISSUE #479: Convert component: components/character_select.vue

AriannaAltomare:
(Disclaimer: I am a Vue noobie and am relying on migration documents and other examples I've found. Please validate this work carefully since this isn't my immediate domain of expertise 😉 ).

Converts a simple Vue component (character-select) to the new Composition API - using this as a quick trial to make sure I'm on the right path here, before converting other simple components and moving on to much more involved migrations. Testing seemed to work normally for this element sans the default not being set (looking into now) but would love more eyes to test and find flaws.

Resolves #479

### DIFF
--- a/components/character_select.vue
+++ b/components/character_select.vue
@@ -1,5 +1,5 @@
 <template>
-  <select v-model="value" @change="onSelectChange" class="form-select">
+  <select :value="selected" @change="onSelectChange" class="form-select">
     <option
       v-for="character in characters"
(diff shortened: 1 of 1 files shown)


AriannaAltomare:
Seems to work in all spots now properly (kink compare, friend add, guestbook post) - should be a low-risk issue regardless.

PR #594, opened 2026-01-08
Implement proper embedding for Bluesky links

### LINKED ISSUE #579: Make Bluesky embeds use FxEmbed too
**Is your feature request related to a problem? Please describe.**

This was briefly mentioned in the Discord, but Bluesky embeds do not use any special parsing. Which in our case means that the image previewer often shows the profile pic for its image. 

Twitter links use FxEmbed’s API to fetch the images and videos, which also supports Bluesky links. It won’t be too much trouble to also make this work for Bluesky then. 

https://github.com/FxEmbed/FxEmbed

freenutsxd:
FxEmbed (and every other embedder that I tried) doesn't play nicely with Bluesky videos in Horizon, so instead we're using [VixBluesky](https://github.com/Lexedia/VixBluesky) which is the only solution I've found that renders videos properly. 

Unfortunately, we either have to choose between videos working OR having a working gallery view for posts with multiple images, but I figured that working videos mattered more. We don't even show gallery views for Twitter posts, anyways. 

Closes #579

### DIFF
--- a/chat/preview/image-url-mutator.ts
+++ b/chat/preview/image-url-mutator.ts
@@ -109,7 +109,8 @@ export class ImageUrlMutator {
       async (url: string, match: RegExpMatchArray): Promise<string> => {
         const path = match[1];
 
-        return `https://d.fxbsky.app/${path}`;
+        // https://github.com/Lexedia/VixBluesky/wiki/Features#custom-pds-video-support
+        return `https://r.v.bskx.app/${path}`;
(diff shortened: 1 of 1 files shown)


FatCatClient:
That makes sense. I definitely remember having a hard time getting videos to open in a new tab on BlueSky, so I'm not surprised that FxEmbed struggled to implement something for that themselves.

PR #600, opened 2026-01-12
Fix for .focus() on alt+tab

snowsune:
This isn't an ideal solution, it just brute-force checks if you were using alt before switching.

A better solution would swap out when we run the .focus and do it only for clicking on a notification box or optionally receiving a message.

Currently
```js
// notifications.ts
browserWindow.webContents.send('request-focus');

// Window.vue  
electron.ipcRenderer.on('request-focus', () => {
  // handle focus
});
```

Maybe instead of calling that request focus in there we could listen for the ipc from clicking, or whatever.
Then trigger a slightly smarter `do-focus` somewhere else?

### DIFF
--- a/electron/Window.vue
+++ b/electron/Window.vue
@@ -430,18 +430,20 @@
           this.show(this.tabMap[id]);
         }
       );
+
+      // Handle explicit focus requests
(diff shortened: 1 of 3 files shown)


snowsune:
Thank you very much

PR #602, opened 2026-01-15
Automatic updates in production

CodingWithAnxiety:
Hello, maintainers of this project! Come haz a look at my cool new PR.

Anyways, joking aside. This PR is a big one, apologies in advance, @FatCatClient .

## Changes
- Adds update checks when the build is prod, with a settings toggle and beta channel setup. (Haven't tested beta stuff, sowwy.)
- Shows update availability thru the UI, with a changelog. Also allows you to update from this dialogue. So you know, it's a little more.
- Release notes are handled differently and now each release gets a doc under docs/releases. PRs should write their changelog updates to this.
   - the bump_version.sh script should automatically generate changelog.md from the release notes in `docs/releases`. A `scripts/changelog-tools.mjs` tool also provides tools for regenerating `docs/releases` stuff or `changelog.md` as needed.  
- CI will pull from the release notes for that tag, now, and attach it to the draft release.

@FatCatClient is requested to review this. I've also pulled copilot in to help catch stuff we don't. On that note, I'd like to start using copilot more to catch small mistakes we might make and not catch, I think that's a fair use of AI to save us all some time? :emu_shrug:

k thx bye ♥


FatCatClient:
Here's what I think we should do before we can merge into development:

- [x] Fix merge conflicts (obviously)
- [ ] Streamline some of the confirm and popup dialogs during the whole 'Update becomes available' > 'User confirms install' process. There are a few too many of these, and I think they're a bit unnecessary
  - Ideally it's just: 
    - "Button to download appears"
    - User clicks button and dialog shows up _(alternatively no user input required for that dialog to appear but it may be annoying)_
    - User confirms in dialog they want to download update
    - Update downloads in background
    - When done, the user can click the install update button (which replaces the update available button) to restart and apply the update. Prompting them only if there are active and connected tabs
- [ ] Add visual feedback to the main window that an update is being downloaded after the user requests a download. Slower internet speeds mean that people need to know an action is still taking place, that way they don't think the system isn't working when downloading a 200mb file.
  - I was personally just thinking of a spinning circle loading indicator where the update button normally appears, but we can see
- [ ] Implement the automatic download functionality. Ie the checkbox in the update prompt (and in the app settings window) that skips all of the prompting and instantly shows the restart button on top.
- [ ] If an update is already downloaded but not installed when the app is closed, don't make it first open a window for two seconds and _then_ restart for the update. This feels weird and janky.


@CodingWithAnxiety

subtervisor:
Are you going to be signing/notarizing the macOS builds now? That's the only way for proper automatic update to work, otherwise you run into the same issue I hit now when upgrading (having to re-grant keychain access and type in my login password). Since the keychain uses code signing identity, the current builds all show up as a different app identity, triggering this behavior.

FatCatClient:
> Are you going to be signing/notarizing the macOS builds now? That's the only way for proper automatic update to work, otherwise you run into the same issue I hit now when upgrading (having to re-grant keychain access and type in my login password). Since the keychain uses code signing identity, the current builds all show up as a different app identity, triggering this behavior.

@subtervisor 

Yes, and acquiring that certificate was one of the things that held this PR back for longer than we had hoped.

Strictly speaking, signing is straight up a requirement for the updater to even work. Not just for the annoying permissions that need to be regranted, so that's why it has stalled for so long.

Soon™️

PR #604, opened 2026-01-19
Feature: New chatbox

CodingWithAnxiety:
Adds a new chatbox to the Horizon interface.

Notably, this layout defaults to the buttons being under, and also is generally a little neater. Still needs some testing. PR as draft.


FatCatClient:
Can we get screenshots, please?

I think that playing with custom font sizes for everything doesn't help in making it fit in with the rest of the UI. Don't we have standard classes for most of those?

CodingWithAnxiety:
Closing as stale: Maintaining the branch as a reference for future work.

PR #640, opened 2026-02-18
Global chat/ character settings

### LINKED ISSUE #190: Set Horizon Settings as Default
It's getting to be a bit of a chore to constantly go and set the Horizon-specific options for every single character I have.  Can we get a button on that tab that just sets the currently checked options as the default option regardless of character you're on?

### LINKED ISSUE #458: Can't read logs until I've signed on as a character
I'm using the OS X version. The latest update brought back an old issue from the Rising days that had previously been fixed, where logs won't load until I've actually been on the chat as at least one character. Meaning I can't just open the client and look at logs, I have to actually open the chat at least once, even if I don't actually want to talk to anyone. Doesn't matter what character, none of them display and then when I do go on chat they all display properly again until I restart the client.

Again, this was an issue that Rising and the early days of Horizon had but was subsequently fixed the first time a while back, and is now rearing its ugly head once more with 1.34.0

FatCatClient:
TODO:

- [x] Bring up to date with post-options API changes in 2.0.0
- [x] 2.4 locale changes
- [x] Version number for settings file(s)
- [x] Proper system for messaging other tab instances that they need to reload the global settings
  - do we use the ``configuration-updated`` eventbus event for this?)
- [x] ~~Figure out how we want to deal with other "settings", like favourite eicons, hidden users, etc.~~
  - Out of scope for this PR. We might do that later with a unique UI solution.
- [x] Make the SettingsView UI not suck.
  - [x] Right now the SettingsTriView component and the logic to switch between modes is an awful hack. Let's work with a component that encompasses the whole boolean/ tri-state setting and uses the settings mode as an input.
  - [x] Only 2 settings actually make a distinction between the two modes. But I'll redo it when I have a cleaner method.
  - [x] ~~Some settings have their checkboxes enabled/ disabled based on other states. This does not have an equivalent with the new parent component(s).~~
    - This does work, but it just needs to be applied to different places.
  - [x] The actual design for the component will need to be redone with vertical tabs too.
- [x] Smart filter menu stuff. Ughhhh.
  - tl;dr: The only changes I've done for that menu is to make sure that the functions to save/ load don't completely break because of the way these are written. They VERY likely don't work at all now and might need a rewrite to an extent.
- [x] The tabs component had to be significantly rewritten for some of the new features, but the code to save me time to not have to update the API for every single other instance of the component is awful. Let's clean that up later.
- [x] Move SettingsView component to a child of Index.vue, instead of ChatView.vue
- [x] Clean up SettingsView code to use a singular Settings object instead of a million fields for every individual setting.

Closes #190.
Closes #458.

Stuff to do AFTER merging (to be made into their own issues), but we can live without these during the initial dev builds:

- ~Make SettingsView handle app-level settings too:~
  - ~Move UI elements over~
  - ~let settings menu save the checkbox type settings~
  - ~add event handlers for browse stuff~
  - ~(optional) theme preview box for live previews~
  - See #904 
- ~Replace placeholder discord icon in SettingsView with custom horizon font icon~
  - Resolved on branch already.
- ~~Initial import dialog issues~~:
  - ~~'Cancel' button does not work, I forgot that this needs custom logic for the Modal component (and it doesnt have custom text either)~~
    - Fixed in branch
- Importing your _character_ overrides to the new format doesn't work 100%: it does not take into account whether your pre-2.4 settings match your 2.4 global defaults, and will set them to their _real_ values instead of ``undefined`` even if they match
  - This might need to be done in two instances: the initial import dialog and loading other character settings.
- ~Places that check your character directories (import settings, log viewer) need to ignore the "_" character directory~
  - Resolved on branch already.
- Fix settings importer (between characters)
  - This hasn't been tested yet, but it _might_ be broken.
- ~Add checkboxes to settings importer to pick individual files to import (main settings, hidden users, pinned eicons, etc)~
  - Out of scope. Let's make this a feature at a later date.


chat/SettingsView.vue
```diff
@@ -1277,88 +1906,28 @@
...
-        this.hideNonCharacterFriends = settings.hideNonCharacterFriends;
-        this.bbCodeBar = settings.bbCodeBar;
+        const globalSettings = core.state.globalSettings as any;
+        const defaultSettings = new DefaultSettings();
+
+        for (const [key, value] of Object.entries(defaultSettings)) {
+          if (globalSettings[key] === undefined) globalSettings[key] = value;
+        }
+
+        this.settings = { ...globalSettings };
```
freenutsxd, line 1906: `this.settings ={ ...globalSettings, risingFilter: JSON.parse(JSON.stringify(globalSettings.risingFilter))};`

This makes a copy of the settings so that when a user changes a filter, rebuildFilters() can see there's a change to retroactively hide/show ads

chat/SettingsView.vue
freenutsxd, line 2091: We should retarget these to `this.settings.risingFilter`

chat/SettingsView.vue
```diff
@@ -1155,10 +1825,18 @@
...
   import core from './core';
-  import { Settings as SettingsInterface, Character } from './interfaces';
+  import {
+    Settings as SettingsInterface,
+    Character,
+    PartialSettings
+  } from './interfaces';
   import l from './localize';
   import {
     SmartFilterSettings,
```
freenutsxd, line 1824: This import isn't needed anymore, see other comments

chat/SettingsView.vue
```diff
@@ -1179,84 +1857,35 @@
...
-        risingFilter: {} as any as SmartFilterSettings,
+        characterOverrides: {} as PartialSettings,
 
-        horizonPersistentMemberFilters: undefined as any as boolean,
         risingAvailableThemes: [] as ReadonlyArray<string>,
-        risingCharacterTheme: undefined as string | undefined,
-
         smartFilterTypes: smartFilterTypesOrigin,
+        settings: { ...defaultSettings } as DefaultSettings,
+        risingFilter: {} as any as SmartFilterSettings,
```
freenutsxd, line 1867: Not needed if we retarget lines 2091-2103

freenutsxd (dismissed):
A few tweaks following my changes in #826. Everything else _should_ work fine

FatCatClient:
@CodingWithAnxiety 

I think it's now in a state where I'd feel comfortable merging it into the dev branch. There are still a handful of tasks that need to be done (see my list of bullet points and the two issues I've preemptively made for some of them). 

But as-is, none of the existing functionality has regressed. I just felt that if we merge it before doing those things, then we don't have to deal with holding up other things because of this PR. 

Aside from moving stuff from Settings.vue to chat/SettingsView.vue (which is more so tedious than hard), most of the items on that list are fairly small bite-sized chunks anyway. I can elaborate more on them in their own Github issues once this is merged.


Give me the go ahead on merging this, if you agree with me.

PR #679, opened 2026-03-04
Log viewer loading fixes

### LINKED ISSUE #648: Removing search terms dumps you at the top of the loaded section of the log no matter what you are currently scrolled to.

### LINKED ISSUE #649: Scrolling up the logs can still be a bit jittery. Minor issue as it's actually possible to scroll up the logs now.

### LINKED ISSUE #652: Jumping to a message does not always consistently load the surrounding messages. Scrolling up and down a bit will properly load the rest.

### LINKED ISSUE #655: Occasionally swapping dates can cut off the bottom-most message and make it impossible to scroll to see the full message.

### LINKED ISSUE #656: Opening logs in general does not always show all the messages it can; a quick scroll up will fix this.

### LINKED ISSUE #666: Switching between dates in the log viewer puts you at the first message, instead of the most recent one
### Describe the bug

The vanilla (and current behavior on the stable release) puts you right on the bottom of a conversation when you pick a date. Right now it puts you on top.

This only happens when you pick a specific date from the third dropdown. Switching to a different conversation altogether puts you on the bottom like it should.

### Version Information

Version: 1.36.0-beta.2
Commit: 0f681c2
Electron: 39.2.7
Chromium: 142.0.7444.235
Node.js: 22.21.1
Platform: macOS ARM64 (25.3.0)

### To reproduce

1. Open a conversation in the log viewer
2. Pick a specific date that's long enough to be scrollable
3. Be put on top instead of on the bottom

### Additional context

This is very likely similar or downright the same bug as #648.

freenutsxd:
Fixes for a handful of the log loading issues that came with 1.36.0-beta.1. Many of the major issues stemmed from container heights not being adjusted/read properly, and the visible range of messages not being updated as frequently as it needed to be. In addition to the issues below, I also found _another_ bug with logs specifically related to resizing/minimizing the window, which has now been fixed.

Currently fixes #656, fixes #655, fixes #649, fixes #648, fixes #652, fixes #666 <sub>thanks github, this looks really cool</sub>

### DIFF
--- a/components/VirtualList.vue
+++ b/components/VirtualList.vue
@@ -232,7 +232,13 @@
         this.programmaticScroll = false;
         return;
       }
-      this.scrollLockedToBottom = false;
+      const el = this.scroller;
+      if (el) {
+        this.scrollLockedToBottom =
+          Math.abs(el.scrollHeight - el.scrollTop - el.clientHeight) <= 1;
+      } else {
+        this.scrollLockedToBottom = false;
+      }
       if (this.settleTimer !== undefined) clearTimeout(this.settleTimer);
       if (!this.scrollbarHeld) {
         this.settleTimer = setTimeout(() => {


freenutsxd:
> Doesn't seem like that's completely the case. It's fine when I use the classic view, but when I enable modern view for messages it still gets cut off. This only happens when viewing logs for the console tab though.

_Now_ it should be fixed in commit 16abbda

PR #692, opened 2026-03-13
feat: Add a friend list to the chat UI

CodingWithAnxiety:
This ads a friends list to the chat UI.
Note that this is NOT complete. We should probably update it to use virtuallist.vue for preformance, though it has been moderately stress tested.


FatCatClient:
Copy pasting my list of points I sent on Discord to here, so that I don't have to keep the text file myself anymore:

```
- friends menu should probably be modal dialog instead of conversation item. replaces sign out button on the left, which gets moved and shrunk upward
- seeing your ignores and hidden users is good. means we can get rid of those things in the settings menu maybe
- filters and sorting is really good
- showing who you're friended with below a friend is really good
- the all friends/ bookmarks lists should probably use a virtuallist component. theyre pretty laggy if the list is large
- should the friend remove dialog have "Show all" if you're only friends on one person?
- bookmark/ friend remove dialogs reset their value on the close event, instead of the afterClose (don't know from the top of my head) event. this means you can visibly see it reset while the animation plays. status bbcode dialog had the same issue before, so just yoink the fix from there

The maybes:
- I don't know if having two rows next to each other all the time makes it cluttered or not. my gut feeling says yes, but I don't know how other people feel about it. let people test that
- using the shift button to remove friends is going to lead to some mistakes
- using the shift button to remove a bookmark might be better if the bookmark itself remains there (with the bookmark button reversed?). means you can undo it easier if you do it by mistake
```

CodingWithAnxiety:
Closing with intention of a 'do-over'

PR #712, opened 2026-03-27
Tweaks for mobile-size quick switcher on top:

### LINKED ISSUE #698: The narrow view's conversation switcher lacks some features
### Describe the problem

When viewing the app in a narrow view (ie the 'mobile' one), a list of your conversations appears on the top of the window for easy switching. While they're obviously stripped down from the main conversation list items in the sidebar, they lack some interactions that I think would be good to have and wouldn't clutter things up. Those being:

- Notification count badges (2.0.0 and later)
- Middle mouse button to close a conversation
- Drag and drop reordering (debatable. it might be annoying for touch inputs)
- Right clicking DMs to open the user menu

### Your idea/ solution

Just add the same stuff to these ones as well.

### Additional context

I'm tagging this partially under "Pre-release build" because one of the items that's missing (the notification badges) aren't in the stable release yet.

FatCatClient:
- Mouse controls (middle and right click match the lefthand side)
- Layout centered correctly
- Visible unread badges
- Setting to forcefully enable

Closes #698

### DIFF
--- a/chat/ChatView.vue
+++ b/chat/ChatView.vue
@@ -278,7 +278,11 @@
         padding-bottom: 10px;
       "
     >
-      <div id="quick-switcher" class="list-group">
+      <div
+        id="quick-switcher"
+        class="list-group"
+        :class="{ forced: forceQuickConvoList }"
+      >
         <a
           :class="getClasses(conversations.consoleTab)"
           href="#"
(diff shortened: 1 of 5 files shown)


FatCatClient:
I accidentally committed changes to that .scss file because I needed some of those helper classes earlier. Whoever wants to click that merge button, don't do it until I force-push a version that doesn't have that change.

Issue #581, opened 2025-12-28
Friends can sometimes appear twice (and not be removed from the online list properly)


FatCatClient:
**Describe the bug**
Reported in the Discord.

Occasionally a friend might appear twice in the friends list. When that person then logs off, one of the names will remain in the list with their status icon being turned into the offline X.

**Screenshots**

<img width="23" height="333" alt="Image" src="https://github.com/user-attachments/assets/456d20c8-28ae-4f38-a9a9-49018160b5ac" />

<img width="204" height="359" alt="Image" src="https://github.com/user-attachments/assets/337a8074-426b-45f8-b519-034447dbaaba" />

**Version and system info:**

- OS: All.
- Version 1.35.4

**Additional context**
We don't have a direct way to reproduce this (yet), but the consistent factor seems to be that in both screenshots the friend in question seems to be in the "Other" list.

Another exchange from the Discord gives us a good lead too:
> you're not friends with that same person across multiple profiles, are you?

> oh! I was not, and then they friend requested the profile I was logged in on. So maybe that's the source of the error?

FatCatClient:
Reproducible case discovered:

1. Be friends with profile X on a different profile than the one you're on
2. Have profile X friend request you, either before or after
3. Accept the friend request
4. Profile X now appears twice in your "Other friends" list

Relogging fixes the issue too.

PR #725, opened 2026-04-02
Ancient issue fixes

### LINKED ISSUE #109: Character colours aren't updated in the recent conversations tab
Ancient F-List 3.0 bug, but might as well fix it ourselves.

Reproduction:

- Have a user in your recent conversations menu be offline
- Wait for them to log in

Their name remains the same dulled grey colour as if they were offline.

### LINKED ISSUE #112: Notifications should ideally show little to no BBCode tags
As is, notifications still have the 3.0 behaviour where they just dump the raw message text in the notification body, which doesn't help make them more readable-- especially for messages with lots of tags (like coloured, bold sub-tagged text). Keeping spoiler tags hidden and eicon names still visible would be the ideal solution too, it's similar to what Discord does with spoilers and custom emoji in its notifications.

The "Copy without BBCode" option in the right-click menu would have been a useful thing to look at, were it not for the fact that it doesn't actually look at the selected BBCode and simply copies the text in the HTML.

### LINKED ISSUE #143: Userview right click menu is (partially) case sensitive

freenutsxd:
Fixes for a few ancient issues on the repo. 

- The bookmark button when right clicking a user is no longer case sensitive.
  - cda8da91ab29c78856cdb525c23c54bfd280aaa6 fixes #143 
- The recent conversations view will update a user's gender color properly when they log in. 
  - In my testing, I noticed that Horizon color is exempt from this logic, oddly.  Characters with Horizon colors will show their custom colors at all times, _even when offline_. Do we care enough to make it consistent?  
  - d096b44f6a347b4b733fa5cc005c84f1a8fec2e9 fixes #109 
- BBCode is stripped from messages for OS notifications. Eicons are surrounded with `:` a la Discord and spoiled text is actually hidden. 
  - 3d525fcfaf840eddcc5312242035d0a198b2470e fixes #112 

notif

### DIFF
--- a/fchat/characters.ts
+++ b/fchat/characters.ts
@@ -184,9 +184,10 @@ export default function (this: void, connection: Connection): Interfaces.State {
       };
     for (const key in state.characters) {
       const character = state.characters[key]!;
-      character.isFriend = state.friendList.indexOf(character.name) !== -1;
-      character.isBookmarked =
-        state.bookmarkList.indexOf(character.name) !== -1;
+      character.isFriend = state.friendList.some(f => f.toLowerCase() === key);
+      character.isBookmarked = state.bookmarkList.some(
+        b => b.toLowerCase() === key
+      );
       character.status = 'offline';
       character.statusText = '';
     }
--- a/fchat/characters.ts
+++ b/fchat/characters.ts
@@ -73,7 +73,7 @@ class State implements Interfaces.State {
       char.isBookmarked = this.bookmarkList.indexOf(name) !== -1;
       char.isChatOp = this.opList.indexOf(name) !== -1;
       char.isIgnored = this.ignoreList.indexOf(key) !== -1;
-      this.characters[key] = char;
+      Vue.set(this.characters, key, char);
     }
     return char;
   }
--- a/fchat/characters.ts
+++ b/fchat/characters.ts
@@ -69,8 +69,8 @@ class State implements Interfaces.State {
     let char = this.characters[key];
     if (char === undefined) {
       char = new Character(name);
-      char.isFriend = this.friendList.indexOf(name) !== -1;
-      char.isBookmarked = this.bookmarkList.indexOf(name) !== -1;
+      char.isFriend = this.friendList.some(f => f.toLowerCase() === key);
+      char.isBookmarked = this.bookmarkList.some(b => b.toLowerCase() === key);
       char.isChatOp = this.opList.indexOf(name) !== -1;
       char.isIgnored = this.ignoreList.indexOf(key) !== -1;
       Vue.set(this.characters, key, char);
--- a/electron/notifications.ts
+++ b/electron/notifications.ts
@@ -17,6 +17,11 @@ export default class Notifications extends BaseNotifications {
   ): Promise<void> {
     if (!this.shouldNotify(conversation)) return;
     this.playSound(sound);
+    body = body
+      .replace(/\[spoiler\].*?\[\/spoiler\]/gi, '██████')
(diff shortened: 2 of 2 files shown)


FatCatClient:
Freenuts always delivers, bless. 🙏 

> In my testing, I noticed that Horizon color is exempt from this logic, oddly. Characters with Horizon colors will show their custom colors at all times, even when offline. Do we care enough to make it consistent?

Yep. iirc, it also does this for things like the message-view component. I don't know how this would be handled with perfect consistency w/ gender colors, since those get set to ``none`` (as if the character was genderless) if you try and check the gender for a character that hasn't logged in at all during that session. Only if the relevant setting that completely removes the color after logging in is enabled, is the color an accurate reflection of the online state.

If you want to make it consistent (which is probably something we have to check is a good idea or not), I'd make it so that the custom color also disappears when "Make color in chat change to grey when user goes offline" is turned on.

On a similar note, I'd also make it so that the gender symbols you can toggle in a UserView component work the same way. Right now they're always set to the X symbol if the user goes offline, even if we have an actual gender stored. This doesn't really make sense to me, if it's meant to be in part an accessibility feature.

freenutsxd:
I was thinking it'd probably be better to make custom colors more consistent and follow the same `none` coloring rules as gender colors, but I think that's something we can ask people about later on in case anybody _really_ likes their custom colors while offline. Same thing with the gender markers, in case any colorblind users would prefer having the X marker so they can tell who's offline rather than the accurate gender being shown at all times. Either way, I'd be happy to work on some tweaks for both if people agree

FatCatClient:
It's probably out of scope for this PR (unless you're really bored), but I do think that the gender marker should 1:1 match with the color that's being set for the gender. Including taking into account the setting that removes the gender color after logging out, rather than _always_ doing that behavior.

If they want the X marker, then wouldn't it make more sense if it shows up when that specific setting is enabled? Otherwise (if you don't have the setting enabled) you lose that information if someone logs off and you're relying on gender markers.

fchat/characters.ts
```diff
@@ -69,11 +69,11 @@ class State implements Interfaces.State {
     let char = this.characters[key];
     if (char === undefined) {
       char = new Character(name);
-      char.isFriend = this.friendList.indexOf(name) !== -1;
-      char.isBookmarked = this.bookmarkList.indexOf(name) !== -1;
+      char.isFriend = this.friendList.some(f => f.toLowerCase() === key);
```
FatCatClient, line 72: Does this also resolve #581?

CodingWithAnxiety (commented):
can copilot please not review on my behalf kthx bye

fchat/characters.ts
CodingWithAnxiety: I don't believe so, I think 581 is a separate race condition but I very well could be wrong.

PR #726, opened 2026-04-03
fix: correct image gallery loading state order

Kannamoris:
`show()` function was loading images from cache but leaving `loading` as true. Alsoo removed the redundant `loading = true` in `showAsync()` that was re-hiding cached images during the background refresh.

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -450,6 +450,17 @@ async function onReady(): Promise<void> {
           return;
         }
 
+        if (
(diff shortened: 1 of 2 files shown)


FatCatClient:
Cheers for this PR. I noticed that last issue you mentioned earlier, but hadn't taken the time to investigate it myself yet.

CodingWithAnxiety:
This looks fairly solid, merging. Thanks for your contribution!

PR #734, opened 2026-04-16
Newline consistency fixes

### LINKED ISSUE #727: Newline being wrongly added before '/me' command when two eicons are in message
### Describe the bug

When attempting to send a message using the `/me` command with two eicons, like this:
```
/me [eicon]foxdisappear[/eicon]
[eicon]foxappear[/eicon]
ignore just testing a thing
```
The message seems to add a extra newline that isn't on the body of the typed message
Image

This causes F-Chat not to interpret the `/me` command, and send as if it s a normal message, which is not the desired outcome.

### Version Information

Version: 2.0.2
Commit: 349f4279
Electron: 39.5.2
Chromium: 142.0.7444.265
Node.js: 22.22.0
Platform: Linux 64-bit (7.0.0-rc5-1-mainline)

### To reproduce

1. Go to anywhere that can be sent a message on chat
2. Send a message with two eicons one underneath the other, like:
```
/me [eicon]foxdisappear[/eicon]
[eicon]foxappear[/eicon]
```
3. The message will have a extra newline before the `/me` command that wasn't there before

### Additional context

I believe this is related to the whole eicon grid feature, I remember(could be misremembering) there is a feature where if you add a 'grid' of eicons (2x2 or 3x3), it'd automatically add a newline so that the grid wouldn't be broken
Maybe it is wrongly identifying a grid and triggers the newline?

Beel, a friend of mine, authorized to share the screenshot of our chat where this was discovered for the bug report:
Image

### LINKED ISSUE #728: Newline characters are escaped for the character preview
### Describe the bug

``[icon]`` and ``[user]`` tags with a newline inside them, have this character ignored (as does the right-click UserMenu). Ie, ``[icon]Mister\n Cool[/icon]`` produces the same icon element as ``[icon]Mister Cool[/icon]``. However, the hover preview on the top left escapes the name at the first ``\n``, meaning that it would instead show for the character "Mister".

### Version Information

Version: 2.0.3-beta.1
Commit: 43c4ee7
Electron: 39.8.3
Chromium: 142.0.7444.265
Node.js: 22.22.1
Platform: Windows 64-bit (10.0.22631)

### To reproduce

_No response_

### Additional context

_No response_

freenutsxd:
Fixes for some recent newline-related issues that were opened.
- Newlines within `[user]` and `[icon]` tags no longer break the character preview.
  - 6ff32c78877eb03b601809843fe2e37c9bd95b3e fixes #728  
- `/me` messages will now prepend newlines for eicon mosaics correctly. 
  - The fix done for the modern chat view will also make it so that `/me` messages with manually prepended newlines (e.g. user hits shift+enter themselves) will now show the newline. _Personally_ I think this is fine since the current implementation doesn't even add a space after stripping the newline, so it just reads as `*[Character]jumps`
  - f6f50e42b729284ff8ef2ec08d0dda400f3af8ae and 6e2eb216851ee77024264a62c9021841953df425 fixes #727  

image 

This was sent with just the following: 
```
/me [eicon]foxdisappear[/eicon] me 
[eicon]foxappear[/eicon] you
```

### DIFF
--- a/bbcode/IconView.vue
+++ b/bbcode/IconView.vue
@@ -48,7 +48,7 @@
     },
     methods: {
       getCharacterUrl(): string {
-        return `flist-character://${this.character.name}`;
+        return `flist-character://${this.character.name.replace(/[\r\n]/g, '')}`;
       },
       dismiss(force: boolean = false): void {
         EventBus.$emit('imagepreview-dismiss', {
--- a/bbcode/IconView.vue
+++ b/bbcode/IconView.vue
@@ -23,7 +23,7 @@
   import Vue, { PropType } from 'vue';
   import { EventBus } from '../chat/preview/event-bus';
   import * as Utils from '../site/utils';
-  import { characterImage } from '../chat/common';
+  import { characterImage, normalizeCharacterName } from '../chat/common';
   import { Character } from '../fchat';
 
   export default Vue.extend({
@@ -48,7 +48,7 @@
     },
     methods: {
       getCharacterUrl(): string {
-        return `flist-character://${this.character.name.replace(/[\r\n]/g, '')}`;
+        return `flist-character://${normalizeCharacterName(this.character.name)}`;
       },
       dismiss(force: boolean = false): void {
         EventBus.$emit('imagepreview-dismiss', {
--- a/bbcode/core.ts
+++ b/bbcode/core.ts
@@ -7,6 +7,7 @@ import {
 import * as Utils from '../site/utils';
 import { default as IconView } from '../bbcode/IconView.vue';
 import UrlTagView from './UrlTagView.vue';
+import { normalizeCharacterName } from '../chat/common';
 import core from '../chat/core';
 
 const urlFormat = '((?:https?|ftps?|irc)://[^\\s/$.?#"\']+\\.[^\\s"]+)';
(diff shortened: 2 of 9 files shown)


FatCatClient:
It's not because of any changes you've made yourself (they're vanilla issues apparently), but I'd also check the behaviour of the UserMenu component while you're at it. It seems to be providing _very_ inconsistent results. Both for the name it displays, and the actual name value it's passing along to the functions for the buttons.

For the name:

```
mr\n
test
```
The previewed name will show ``mr test``, the profile pic will be for ``mrtest`` and buttons like the ignore or bookmark/ unbookmark ones will use an actual encoded newline character: ``%0A``.

freenutsxd:
> It's not because of any changes you've made yourself (they're vanilla issues apparently), but I'd also check the behaviour of the UserMenu component while you're at it. It seems to be providing _very_ inconsistent results. Both for the name it displays, and the actual name value it's passing along to the functions for the buttons.

The best idea I could come up with to solve this was to normalize all the character name inputs to strip newlines from them, which seems to work pretty well ([82665f8](https://github.com/Fchat-Horizon/Horizon/pull/734/commits/82665f883b251f828eff54ef14e462ae14f2200b)). I added a separate `displayName` variable that'll hold what's actually displayed in chat, for the very rare case someone wants the formatting below to show up as intended:
```
[user]mr\n           // this still shows as: mr
test[/user]                                  test
```
Now in the example you gave, it'll show `mr test` accurately across all instances, and the UserMenu buttons will also work.

PR #744, opened 2026-04-25
Revert "fix:/ feat: Gender icon now only becomes an X if usernames are set to gray if users are offline (matches gender color at all times)

FatCatClient:
But I'll probably poll it in the Discord before merging this to dev.


CodingWithAnxiety:
Oops, did this without merging the PR ♥

PR #746, opened 2026-05-04
fix: Escape in image preview no longer closes the entire profile

Kannamoris:
Pressing escape on an image preview now closes the preview instead of the whole profile


FatCatClient:
Thanks for the PR!

PR #749, opened 2026-05-05
feat: group system for channels 

Kannamoris:
Channel groups: Adds collapsible channel folders on the left sidebar. Channels can be assigned to named groups via the new channel menu. Groups can also be renamed by double clicking or pressing the delete icon when hovering over it. Group membership persists between sessions, and grouped channels are re-joined on login.

Channel menu: Right-clicking any channel opens up a context menu styled to match the existing character menu, with options to mark as read, copy the channel code and manage group memberships.


chat/interfaces.ts
```diff
@@ -128,6 +128,19 @@ export namespace Conversation {
       character: Character,
       noCreate: boolean
     ): PrivateConversation | undefined;
+
+    channelGroups: Array<{
```
FatCatClient: I think I'd prefer this where an actual typed array instead of defining the properties ad-hoc like this, same with the array on line 270 and the references in the Vue components or conversations.ts. Not a big fan of using ``any`` so eagerly.

Having a single source of truth is good, especially for types that need to be serialized and deiserialized.

chat/ChannelGroupSection.vue
```diff
@@ -0,0 +1,245 @@
...
+        v-model="renameValue"
+        @keydown.enter.prevent="commitRename"
+        @keydown.esc.prevent="cancelRename"
+        @blur="commitRename"
+        @click.stop
+      />
+      <span
+        class="fas fa-fw fa-trash channel-group-delete"
+        @click.stop="deleteGroup"
+        :aria-label="'Delete group'"
```
FatCatClient: For this (and other user-facing strings like the ones in ChannelMenu.vue) could you make sure they're using localized strings?

chat/ChannelGroupSection.vue
```diff
@@ -0,0 +1,245 @@
...
+  const unreadClasses = {
+    [Conversation.UnreadState.None]: '',
+    [Conversation.UnreadState.Mention]: 'list-group-item-warning',
+    [Conversation.UnreadState.Unread]: 'list-group-item-danger'
+  };
+
+  export default Vue.extend({
+    name: 'ChannelGroupSection',
+    props: {
+      group: { required: true as const },
```
FatCatClient: Types for these props, please.

chat/ChannelGroupSection.vue
```diff
@@ -0,0 +1,245 @@
...
+        });
+      }
+      Sortable.create(<HTMLElement>this.$refs['channelList'], {
+        group: { name: 'channels', pull: true, put: true },
+        animation: 50,
+        fallbackTolerance: 5,
+        onStart: () => {
+          document.body.classList.add('channel-dragging');
+        },
+        onAdd: (e: any) => {
```
FatCatClient:
```suggestion
        onAdd: (e: Sortable.SortableEvent) => {
```

chat/ChannelGroupSection.vue
```diff
@@ -0,0 +1,245 @@
...
+          // Channel dropped into this group from ungrouped or another group
+          const channelId = (e.item as HTMLElement).dataset?.channelId;
+          if (channelId)
+            core.conversations.setChannelGroup(
+              channelId,
+              (this.group as any).id
```
FatCatClient: Make sure to remove this (and any other ``any`` casts) if you give ``this.group`` an actual type.

chat/ChannelMenu.vue
```diff
@@ -0,0 +1,200 @@
...
+<script lang="ts">
+  import Vue from 'vue';
+
+  export default Vue.extend({
+    name: 'ChannelMenu',
+    data() {
+      return {
+        showMenu: false,
+        position: {} as { left: string; top: string },
+        conversation: null as any,
```
FatCatClient: As with the other instances of values or variables being un-typed u, please give these an actual type.

chat/ChannelGroupSection.vue
```diff
@@ -0,0 +1,245 @@
...
+          this.$emit('editing-started');
+        });
+      }
+      Sortable.create(<HTMLElement>this.$refs['channelList'], {
+        group: { name: 'channels', pull: true, put: true },
+        animation: 50,
+        fallbackTolerance: 5,
+        onStart: () => {
+          document.body.classList.add('channel-dragging');
+        },
```
FatCatClient, line 123: I love the addition of the little placeholder boxes in empty channel groups, but do you need to add this class to the document's body when the items themselves are in this same component?

If it was also meant to work this way for ungrouped channels, I'd rather those be moved to the same component as well so that all of this logic is compartmentalized

chat/conversations.ts
```diff
@@ -1,3 +1,4 @@
+import Vue from 'vue';
```
FatCatClient: I briefly yapped about this on Discord, but we should probably redo the ``channelGroupAssignments`` construction into something that doesn't require us to directly call Vue functions in this layer to guarantee reactivity.

FatCatClient (changes requested):

FatCatClient:
A few other UX pointers we've had-- which are mostly just CSS things, but I think would greatly help usability for people:

- It should be a bit clearer where you're dragging items to, ie if it's the last collapse or the ungrouped block
- If you fold a group shut, you should still see the ping counter for conversations in there
- If you make a new group or change a name, it should select the full text inside the input box so you don't have to delete the name before typing

Some ideas for the future, but to keep the scope more reasonable we can talk about these after this gets merged and is released to stable:

- Right click menu for channels has options to move them to different groups, which is cool but it might be better if it was a nested menu.
- A right click menu for the group header, with options to merge with another group, dissolve the group, or mark all channels inside as unread.

But if we want to do those two, we're probably better off making a generic configurable component for custom context menu that chat/UserMenu, your new ChannelMenu and this hypothetical ChannelGroupMenu would use, à la CustomDialog.

FatCatClient:
@CodingWithAnxiety Please check if you have any changes you'd want to be made as well or if we are good to go with this PR.

PR #751, opened 2026-05-08
fix: update eicon button tooltip for clarity

Kannamoris:
Had some people not realize there was a way to insert eicon tags instead of having to manually enter it themselves. Hopefully this makes it a little more clear.


FatCatClient:
Now that you're touching these strings, would you mind moving them to localization strings too? I never realized these are hard-coded and unchanged from the vanilla 3.0 client (aside from the Cmd/ Ctrl button stuff I added ages ago).

Kannamoris:
Updated

PR #777, opened 2026-05-19
fix: deleting channel groups now correctly ungroups and unpins channels

### LINKED ISSUE #766: Deleting a channel group does not unpin channels properly
### Describe the bug

See title.

### Version Information

Version: 2.2.0-beta.0
Commit: e57391f1
Electron: 40.10.0
Chromium: 144.0.7559.236
Node.js: 24.15.0
Platform: macOS ARM64 (25.4.0)

### To reproduce

1. Create group of channels with a custom name
2. Add a variety of channels to this list
3. Delete group with the trashcan icon button
4. Close and reopen Horizon
5. Reconnect as the same character
6. The list of channels you thought you just unpinned is back, but is being imported as if they were from the old format.

### Additional context

I suspect this is just because the 'Delete group' button manually clears out the group, instead of calling the function to remove a channel from a group properly (which would also remove it from the 'pinned' list.

Better to have a single ``removeChannel(...)`` function for groups that _also_ gets called for all channels when you delete a group which handles removing it from 'pins' as well.

### LINKED ISSUE #767: Leaving a channel does not properly ungroup it
### Describe the bug

Before, leaving a channel would properly unpin it, meaning you wouldn't automatically rejoin it when you hop onto that same character again. Leaving a channel while it's in a group does not do this at all, meaning you keep rejoining it anyway.

### Version Information

Version: 2.2.0-beta.0
Commit: https://github.com/Fchat-Horizon/Horizon/commit/e57391f12281b3f422107a395b2e36b7ce2290e7
Electron: 40.10.0
Chromium: 144.0.7559.236
Node.js: 24.15.0
Platform: macOS ARM64 (25.4.0)

### To reproduce

1. Put a channel in a group
2. Leave the channel with the X button or by middle clicking it
3. Disconnect and close your tab
4. Open a new tab and sign in with the same character again
5. You will join the channel you thought you removed earlier anyway.

### Additional context

I happened to notice this while working on the channel group right click menu (specifically when adding a method to mark all conversations as unread), but the channel ID is just kept in the ChannelGroup's array of channels. Whatever method is called to unpin it should probably also make sure that it gets removed from a group as well.

Kannamoris:
Fixes #767
Fixes #766

### DIFF
--- a/chat/conversations.ts
+++ b/chat/conversations.ts
@@ -863,11 +863,23 @@ class State implements Interfaces.State {
     return id;
   }
 
+  private removeChannelFromGroups(channelId: string): void {
+    for (const g of this.channelGroups) {
+      const i = g.channels.indexOf(channelId);
+      if (i !== -1) {
+        g.channels.splice(i, 1);
+        return;
+      }
+    }
+  }
+
   deleteChannelGroup(id: string): void {
     const idx = this.channelGroups.findIndex(g => g.id === id);
-    if (idx !== -1) this.channelGroups.splice(idx, 1);
+    if (idx === -1) return;
+    this.channelGroups.splice(idx, 1);
     this.channelGroups.forEach((g, i) => (g.order = i));
     void this.saveChannelGroups();
+    void this.savePinned();
   }
 
   renameChannelGroup(id: string, name: string): void {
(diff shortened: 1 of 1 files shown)


FatCatClient:
Tested it and looks good to me.

PR #795, opened 2026-05-27
Gallery zoom percentage tweak

freenutsxd:
Coming from how image viewers/editors typically work, I was personally thrown off by the way that the full window height in Horizon's gallery counts as 100% zoom, even if the image definitely isn't scaled to 100%. 

This tweaks it so that the zoom percentage is the actual image scale percentage, additionally letting you zoom out to 100% on small images. Large images have their minimum scale percentage set to the window height (i.e. 75%) so that you'll always see the full thing at minimum zoom.

electron_tUrlsAzExs
mspaint_YBvhdsViFq

### DIFF
--- a/scss/_character_page.scss
+++ b/scss/_character_page.scss
@@ -530,10 +530,12 @@
         .zoom-number {
           flex-grow: 0;
           max-width: 20%;
+          min-width: 4.8em;
+          justify-content: center;
           display: none;
           font-size: 0.9em;
           @include media-breakpoint-up(xl) {
-            display: inline-block;
+            display: flex;
           }
         }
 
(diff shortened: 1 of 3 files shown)


FatCatClient:
When switching between images (or rather while the blurry preview is still visible), the zoom percentage is set to 100%, which means that you will now see it jitter between 100 and X% aggressively when going through different images.

Could you replace the zoom slider with a throbber (like what's used for the profile viewer's columns themselves) so that the transition between images isn't as jarring?

FatCatClient (changes requested):
See comment

freenutsxd:
Used Bootstrap's placeholders class to handle the jittering instead of the other throbbers we've used
electron_j1O4RdjuwV

PR #804, opened 2026-05-30
feat: Always open links in incognito toggle

### LINKED ISSUE #525: Feature Request: Open as incognito setting
I feel like a setting to toggle that'd be quite nice is being able to open profiles and links automatically as incognito!

freenutsxd:
Adds a toggle underneath the Browser Path settings to always open links in incognito/private by default when left-clicking them. When enabled, users can still right click any link and select "Open in browser" to open it in a non-incognito tab. The setting is hidden for Mac users since we don't have a good way of opening in incognito on that platform.

Similarly, any profiles opened in browser via the button in the profile viewer, and any gallery images opened in browser in the same way, will also default to an incognito tab.

Closes #525


FatCatClient:
@freenutsxd Could you resolve the merge conflicts before I can test it? If this won't get merged soon-ish, it'll likely have to be after ##904 is done, which would make it even more work for you to get it mergeable.

FatCatClient:
I thought this through for a second, and realized that I'd still rather have this implemented in a way that also refactors the handlers for opening stuff in incognito mode into being called from the main process-- even though I said I'd be fine with doing that refactor later.

Not really sure what reason we'd have for putting it off, since it's mostly a cut and paste job anyway. Other than the fact that it'd clutter up ``electron/main.ts`` even more if none of it is split into its own file.

freenutsxd:
Thankfully it was easier than I thought. Opening in incognito has been refactored to be part of main.ts, now **every** instance where a URL would be opened will respect the 'Always open in incognito' toggle. The only exception to this is manually right clicking and selecting 'Open in browser' on links, since I think that's still nice to have just in case. Tested on Windows and my Linux VM, it seems to Just Work™️.

PR #822, opened 2026-06-12
Refac: Spring Cleaning (General tech debt cleanup)

CodingWithAnxiety:
The general aims of this PR is to help clean up on tech debt to prepare for future changes to the infrastructure for Horizon, ie Vue 3.x / Vite / Eslint / Whatever else. This also helps set a future for more experimental features, IE Sqlite logs.

A big one from this project is the dropping of `@electron/remote`. This includes implementing a IPC interface and setting the following ts flags:

- `contextIsolation: true`
- `nodeIntegration: false`
- `sandbox: true`

It also adds the following:
- A generic preload under `electron/preload.ts`, I may want to clean this up to a similar structure like what we had in Eclipse
- A electron shim, under `electron/electron-shim.ts`, which made my life less painful though webpack aliases. This may, too, want to be cleaned up. But maybe in the future (Future `vite` work?)
- Stuff that needed node, genuinely, and was in the renderer, has been generally moved. 
- Webpack now targets `web` against the renderer, with shims for stuff like `querystring`, `setImmediate`, and `path`

There has also been a dependency cleanup and some refactoring therefore to drop deprecated dependencies. This should hopefully make pnpm nag less about 'vulnerabilities' which don't even affect us.

*This shouldn't be a breaking change, and broken features should be considered... a bug.* This also should be considered a draft, but I'd like to not implement anything too heavy before this is merged.

I'll come up with a better PR description, later, but for now this is good. I am happy.

@FatCatClient, when you're off vacay you **can** take a look at this (not required obv, i'm fairly confident in this), otherwise I'm going to throw it at the hounds (testers) and if it passes well, I'll merge.


CodingWithAnxiety:
Closing with intention of a 'do-over'

PR #824, opened 2026-06-13
Adding a split flake. One for end users, one for devs.

MoonBurst:
I got tired of waiting for a flake, so I tried to make my own. But since there was already a dev flake, I didn't want to muck with it too much. Full disclosure because I still do support intellectual honesty, I did use a bit of the icky in this, But I DID read over the build and test that it builds AND if the dev shell path still builds.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -102,7 +102,7 @@
 
         #horizon-electron for the end user
         #horizon-electron-dev for the devs.
-        packages = {
+        packages = rec {
           horizon-electron = makeHorizon { isDev = false; };
           horizon-electron-dev = makeHorizon { isDev = true; };
           default = horizon-electron;
--- a/flake.nix
+++ b/flake.nix
@@ -1,13 +1,14 @@
 {
-  description = "Horizon dev environment";
+  description = "Nix flakes for both devs and users";
 
   inputs = {
     # why is NixPKGS stable only on Electron 33 as of writing?? (2026-01-06)
-    nixpkgs. url = "github:NixOS/nixpkgs/nixos-unstable";
-    flake-utils. url = "github:numtide/flake-utils";
+    # Electron moved up to 41.7.1 in stable If that matters
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    flake-utils.url = "github:numtide/flake-utils";
   };
 
-  outputs = { self, nixpkgs, flake-utils }: 
+  outputs = { self, nixpkgs, flake-utils }:
     flake-utils.lib.eachDefaultSystem (system:
       let
         pkgs = nixpkgs.legacyPackages.${system};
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -1,13 +1,14 @@
 {
-  description = "Horizon dev environment";
+  description = "Nix flakes for both devs and users";
 
   inputs = {
     # why is NixPKGS stable only on Electron 33 as of writing?? (2026-01-06)
-    nixpkgs. url = "github:NixOS/nixpkgs/nixos-unstable";
-    flake-utils. url = "github:numtide/flake-utils";
+    # Electron moved up to 41.7.1 in stable If that matters
```
FatCatClient: It should be pinned to Electron 40 anyway, so the comment isn't really relevant anymore.

flake.nix
```diff
@@ -16,7 +17,110 @@
...
+            mkdir -p $out/share/applications
+
+            cp -r . $out/lib/fchat-horizon/
+#I don't know if this part is actually needed. I just remember the app image didn't have an icon last time I tried. But I changed some other stuff locally, so I don't know either way TBH
+            find . -name "icon.png" -exec cp {} $out/share/icons/hicolor/256x256/apps/horizon-electron.png \;
+
+            cat << 'EOF' > $out/bin/fchat-horizon
+            #!/bin/sh
+
+            PASSWORD_STORE="detect"
```
FatCatClient: I'd rather this be left out entirely. No other packaged version does it, and it's 100% the kind of thing that would cause problems in the future once the finicky setup changes (or if you run it from an initial install where those daemons are not even running!)

flake.nix
```diff
@@ -16,7 +17,110 @@
...
+
+          buildInputs = [
+            pkgs.stdenv.cc.cc.lib
+            pkgs.libsecret
+            pkgs.glib
+          ];
+
+          desktopItems = [
+            (pkgs.makeDesktopItem {
+              name = "fchat-horizon";
```
FatCatClient: The package and executable name in the official releases (at least the ones i checked) is ``horizon-electron``, so it's probably best to be consistent.

flake.nix
```diff
@@ -16,7 +17,110 @@
...
+              exec = "@out@/bin/fchat-horizon %U";
+              icon = "fchat-horizon";
+              desktopName = "Fchat Horizon";
+              comment = "F-chat desktop client";
+              categories = [ "Network" "Chat" "InstantMessaging" ];
+              mimeTypes = [ "x-scheme-handler/fchat" ];
```
FatCatClient:
```suggestion
              desktopName = "Horizon";
              comment = "The F-Chat Horizon Client";
              categories = [ "Network" "Chat" "InstantMessaging" ];
              mimeTypes = [ "x-scheme-handler/fchat" ];
```

Matches ``electron/build/horizon.desktop``

FatCatClient (commented):
_Maybe_ we should also include some user documentation on how to actually use this flake (and how to get people to pick between stable and unstable releases).

@CodingWithAnxiety do you remember which env var needs to be set to get a built project to use the appropriate updater dialog starting with 2.3.0?

MoonBurst:
Fair. I'll work at it some more, I'll fix these issues, add documentation and get back to you.

CodingWithAnxiety:
> _Maybe_ we should also include some user documentation on how to actually use this flake (and how to get people to pick between stable and unstable releases).
> 
> @CodingWithAnxiety do you remember which env var needs to be set to get a built project to use the appropriate updater dialog starting with 2.3.0?

@FatCatClient, Apologies for being away. I was relaxing for most of yesterday.

As for what env var: It really just depends on what you mean. There's two vars which gate the program's autoupdater on Linux:

- `NODE_ENV`, which should be set to `production` (NODE_ENV=production`) for the autoupdater to appear.
- `APPIMAGE`, which is set on boot of the client. It's a path (IE: `'/home/rose/Downloads/F-Chat.Horizon-2.2.1-linux-x86_64.AppImage'`) if the client is ran from an appimage, and `undefined` if not. This is used by the client to supply the appimage vs package manager autoupdate prompt.

Issue #253, opened 2025-07-02
Smart Filters > Visibility 'hide matching characters' options not working


hedcnist:
**Describe the bug**
Several of the Smart Filter 'Visibility' options do not seem to work. I've only noticed (with varying success) that the 'hide ads from matching characters' seem to work, though specifically in this case I've noticed that 'hide matching characters from search results' and 'hide matching characters from channel members lists' don't.

**To Reproduce**

1. Go to Settings
2. Click on 'Smart Filters' tab
3. Scroll down to 'Visibilty' section.
4. Check 'Hide matching characters from search results' and 'Hide matching characters from channel members lists.'
5. Set a Type Match (let's say Males for this example)
6. Click 'change settings.'
7. Go to any channel you know to have males in it. They will still show in the lists.
8. Search for characters without selecting a gender. Males will still show up.

**Expected behavior**
I assume these, like with their ads, should disappear.

**Version and system info (please complete the following information):**

- OS: Windows 10
- Version: 10.0.19045 Build 19045

FireUnderTheMountain:
My notes for anyone stumbling across this and wants to help:

The simplest fix would be make chat-based filters distinct from profile-based filters and check chat filters explicitly. It would still require a lot of work to implement and be somewhat clunky and have a strange feel both development and user side.

The order of operations to code filter checks in this paradigm is clunky: ProfileCache.getSync() (if sync) -> ProfileCache.get() -> ChatLevelFiltered() (if online) -> core.cache.addProfile() (and watch for response event). You would also need to load *some* kind of profile structure to attach the filtering data onto.

The even simpler "bake-in a fast-check for chat gender to the current filtering" is pointless since filtering/matching only happens as a result of fetching profile data (because filtering mostly depends on reading their profile), and since you've already loaded the data, you can fully check it.

So I've thought a lot about it, and with all the shortcomings to the various ways, I've settled on the idea that the only proper way to implement better filtering is to separate it from profile loading. That is, have a cache/store/api[-equivalent fetcher] that operates parallel to the profile cache/store/api - exclusively for match data. Matcher data is much lighter weight and it would be easier to bulk-fetch data for member lists.

So, if anyone wants to code that, you have my blessings. :)

PR #826, opened 2026-06-15
fix: Smart Filtering + ads consistency

freenutsxd:
Fixes some issues with the smart filter system not working as consistently as it should, causing a bunch of cases where matching characters just don't get filtered at all.

- Having Smart Filters enabled, but the "Colorize ads" setting disabled, wouldn't do anything. Seems like the entire ad scoring system relies on that toggle being enabled, which doesn't feel right imo. Now, specific smart filter settings will still use scoring even if the colorize setting is disabled, allowing filtering and hiding to work as intended, just without the color change.
- The matcher could have a hard time keeping up with a lot of ads coming in at once, resulting in some ads just never getting filtered, so I added some FIFO rules to encourage older ads to be prioritized first.
- Ads will now be shown/hidden retroactively whenever filter settings change. If you change the settings to filter out a gender, previous ads from that filtered gender will now be hidden immediately. 

These changes don't touch the channel member list, so you'll still likely see a lot of characters that _should_ be filtered and/or hidden, but aren't because of the way match checking currently works. I'm not really sure what's a good way of doing that in the first place without fetching the profile of everybody in a channel which sounds like a nightmare, so I didn't touch it.

Although the filters and ad hiding/showing functionality seems to work a lot better in the cases I checked, I'd still love more testing on this from others in case I'm missing something.

### DIFF
--- a/chat/SettingsView.vue
+++ b/chat/SettingsView.vue
@@ -1551,6 +1551,9 @@
             );
           }
         });
+
+        // if a sender's filtered status is stale (not in memory), rematch them
+        void core.cache.rematchStaleAdsInConversations();
       },
       getAsNumber(input: any): number | null {
         if (_.isNil(input) || input === '') {
--- a/chat/SettingsView.vue
+++ b/chat/SettingsView.vue
@@ -1348,7 +1348,7 @@
         this.horizonShowDuplicateStatusNotifications =
           settings.horizonShowDuplicateStatusNotifications;
         this.horizonHighlightUsers = settings.horizonHighlightUsers.join(',');
-        this.risingFilter = settings.risingFilter;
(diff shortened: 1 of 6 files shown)


Kannamoris:
- **Retroactive show/hide on settings change** existing ads from matching characters were hidden after saving filters
- **Filtering with "Colorize ads" disabled** new ads still got matched and hidden
- **Queue** ~80-sender backlog fully drained, nothing stuck unscored

Seems all good to me!

FatCatClient:
@freenutsxd Are there any issues on the issue tracker this PR fixes? I can't find them myself right now, but maybe I missed them.

freenutsxd:
Not really, this was mainly based on a report in the Discord. ~~It _partially_ fixes #253 on the ads side, but the channel member list part wasn't fixed.~~ jk I actually read it and they already said the ads filtering works fine

PR #827, opened 2026-06-16
fix: Tray icon functionality pt. 2

freenutsxd:
Following the changes in #808, @FatCatClient brought up the fact that some people use wholly separate windows for characters as opposed to the standard tabs, which was an oversight on my end. While single clicking to open the app would probably be fine if it was just one window, having the ability to do multiple windows means that single clicking like that isn't going to feel natural.

The changes here tweak the tray icon to now require a _double click_ in order to focus all windows, while left clicking will still show the context menu like before. Additionally, clicking a specific character will now **only** open that character's window as opposed to all of them if they have multiple windows.

### DIFF
--- a/electron/browser_windows.ts
+++ b/electron/browser_windows.ts
@@ -453,7 +453,22 @@ export function createMainWindow(
   if (!tray) {
     tray = new electron.Tray(trayIcon);
     tray.setToolTip(l('title'));
-    tray.on('click', _e => showAllWindows());
+    // single click opens context menu, double click opens all windows
+    let clickTimeout: NodeJS.Timeout | undefined;
+    tray.on('click', _e => {
+      if (clickTimeout) clearTimeout(clickTimeout);
+      clickTimeout = setTimeout(() => {
+        clickTimeout = undefined;
+        tray.popUpContextMenu();
+      }, 200);
+    });
+    tray.on('double-click', _e => {
+      if (clickTimeout) {
+        clearTimeout(clickTimeout);
+        clickTimeout = undefined;
+      }
+      showAllWindows();
+    });
 
     tray.setContextMenu(electron.Menu.buildFromTemplate(createTrayMenu()));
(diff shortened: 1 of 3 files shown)


electron/browser_windows.ts
```diff
@@ -453,7 +453,22 @@ export function createMainWindow(
...
+    // single click opens context menu, double click opens all windows
+    let clickTimeout: NodeJS.Timeout | undefined;
+    tray.on('click', _e => {
+      if (clickTimeout) clearTimeout(clickTimeout);
+      clickTimeout = setTimeout(() => {
+        clickTimeout = undefined;
+        tray.popUpContextMenu();
+      }, 200);
+    });
+    tray.on('double-click', _e => {
```
FatCatClient: I'm pretty sure this event does not fire on Linux. In fact, I think it interprets all click events as the standard `'click'` event. 

It might be better to add a 'Show all windows' item to the context menu as an alternative (though I think it's fine to have that item on all platforms).

freenutsxd: I made it so that the 'Horizon' context menu title is clickable and will show all windows in the last PR, but that might not have been super clear. I could probably change that to say 'Show all windows', since people probably know that they're clicking on Horizon anyways, if you think that'd be better.

Let me see what I can do to get the double click functionality working on Linux

FatCatClient: Nah the Horizon label being there was intentional. Please keep that and just have the show all windows thing be its own item.

freenutsxd:
On Linux specifically, left clicking the tray icon once won't do anything, but double clicking should now work fine. You'll need to right click the tray icon to open the context menu, since my attempts at getting a single left click to open the context menu didn't seem to work.

The context menu was also adjusted, now the 'Horizon' title is no longer clickable, and there's a dedicated 'Show all windows' button (which will need translations)

PR #831, opened 2026-06-16
Quick Switcher visual improvements

freenutsxd:
Quick fix for the mobile-style quick switcher that prevents it from growing in size if someone uses a Horizon portrait that's taller than it is wide. I tried with `object-fit:  cover` at first, but didn't really like how it just displayed a square selection in the middle of the Horizon portrait, so I went with `contain` instead so that nothing is cut out. We can _also_ just make it behave the same as non-quick switcher tabs where it just gets stretched to fit the 30x30 size, so let me know what you guys would prefer.
 
Admittedly I think the quick switcher could use some visual improvements in general, but this is a simple solution until then. 
### Before
image

### After
image

### DIFF
--- a/chat/ChatView.vue
+++ b/chat/ChatView.vue
@@ -1294,6 +1294,7 @@
           img {
             height: 40px;
             width: 40px;
+            object-fit: contain;
             margin: 0;
           }
 
@@ -1349,6 +1350,7 @@
       img {
         height: 40px;
         width: 40px;
+        object-fit: contain;
         margin: -1px 5px -1px -1px;
       }
       &:first-child img,
@@ -1410,6 +1412,7 @@
     img {
       width: 32px;
       height: 32px;
+      object-fit: contain;
       flex-shrink: 0;
     }
 
--- a/chat/ChatView.vue
+++ b/chat/ChatView.vue
@@ -338,7 +338,7 @@
           >
             {{ conversations.consoleTab.unreadCount }}
           </span>
-          {{ conversations.consoleTab.name }}
+          <div class="name">{{ conversations.consoleTab.name }}</div>
         </a>
         <a
           v-for="conversation in conversations.privateConversations"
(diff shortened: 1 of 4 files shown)


freenutsxd:
The more I thought about it, the more I feel like the stretched version might actually look better overall in more scenarios, so I'm leaving this as a draft until a final decision is made. 

**_Also,_** I inspired myself to try and visually improve the mobile quick switcher as well, so I'm gonna convert this PR to focus on that instead. I mainly wanted something that was easier to tap for mobile users and showed just a little more info; see below for my two proposed revisions. 

### 1 (current, just with non-square portraits stretched to fit)
electron_gkx4Y4l2Sx

### 2 (bigger overall, more separated PMs)
electron_jVSDyWr2pB

### 3 (somewhere in between, larger buttons and tap area but still connected)
electron_V1kqIuUquL

freenutsxd:
Option 3 won by a landslide, but I made it just a little bit smaller and added a subtle shadow around the selected tab just for some added depth. It seems to look better and feels easier to use than what we currently have, so until someone complains I think it's good to go. 

### Desktop
Discord_LrLGsTpqsY

### Mobile (Solstice)
image

FatCatClient:
@freenutsxd 

I just happened to randomly notice there are other cases where non 1:1 avatars are also problematic:
- The conversation list items in the left sidebar
- The conversation header on top (where you can also view that conversation's logs, settings, etc)
- the little character icon in message_view if you have it enabled and aren't using Modern view mode

Do you think you could get to those too?


Edit:

> We can also just make it behave the same as non-quick switcher tabs where it just gets stretched to fit the 30x30 size, so let me know what you guys would prefer.

Never mind, you noticed that one too. But that doesn't look particularly good either imo and it'd be better if this solution was brought over to all the above places as well imo

freenutsxd:
I moved every Horizon Portrait occurrence that **isn't** in the profile viewer itself to use `object-fit: contain`, which looks a lot better in practice. Now images won't be squished to fit the 1:1 ratio and will just be smaller as a whole

image

I also made the log viewer's avatar enhancement actually show the Horizon Portrait now, since it was just using the default f-list icons before. But honestly, these images are already so tiny as-is that this might not be worthwhile

image

FatCatClient (approved):

Issue #821, opened 2026-06-11
Client Crashes When Trying To Close Tab


ToxicZombie000:
### Describe the bug

After the 2.2.1 update on Linux, whenever trying to close the client or a tab, it freezes and crashes instead.

### Version Information

Version: 2.2.1
Commit: 4139ac84
Electron: 42.3.0
Chromium: 148.0.7778.180
Node.js: 24.15.0
Platform: Linux 64-bit (7.0.11-1-cachyos)

### To reproduce

Click on the X next to character's name,
Click on the X in the top right to close the window,
Middle click on a tab,
Going through the context menu in the top left, Horizon -> Quit

### Additional context

Any method to log out seems to crash, though I haven't tested all of them. The freeze occurs before the prompt to confirm quitting out shows up.

FatCatClient:
Which version are you using (presumably from the AUR)? The releases _we_ put out are made with and for Electron 40, and not 42.

If the issue doesn't happen in the AppImage or .tar.gz releases, then please take it up with whoever runs that release and make sure it uses the appropriate version of Electron.

CodingWithAnxiety:
Closing as unable to reproduce. If you can reproduce this issue on supported versions of Horizon (Installed from Horizon.moe), please let us know!

Thank you.

Issue #753, opened 2026-05-13
Horizon crashes after attempting to log out (Linux)


j053ph6:
### Describe the bug

When I attempt to log out of a character in Horizon (via the X button or the log out button), the entire app freezes before it eventually crashes.

### Version Information

Version: 2.1.4
Commit: 67331c27
Electron: 42.0.1
Chromium: 148.0.7778.97
Node.js: 24.15.0
Platform: Linux 64-bit (7.0.6-1-cachyos)

### To reproduce

1) Log in with any character
2) Attempt to log out

### Additional context

Running Horizon via the terminal returns this error:

fish: Job 1, 'fchat-horizon' terminated by signal SIGSEGV (Address boundary error)

caderidien:
Definitely an issue related to Electron 42.  Downgrading to Electron 41 fixes the issue but not sure what upstream broke since I don't see any breaking changes that seem relevant.  On Electron 42 and X11 at least I can confirm I had the same issue where you get a memory access violation when Horizon tries to open the prompt that says "You are still connected to chat.  Would you like to disconnect?"

Relmix:
I thought I should add that the same error happens when trying to export logs from within the app. Works with Electron 41 but not Electron 42. Same distro (CachyOS) and versions for everything else.

FatCatClient:
Someone in the Discord was having similar issues on CachyOS because of the AUR version using Electron 42 (instead of Electron 39, which is what we target in the official releases). Though in their case it wouldn't even open a window:

<img width="1608" height="389" alt="Image" src="https://github.com/user-attachments/assets/5e1b8f18-0b17-4f32-a15b-de573cd0f319" />

I think the AUR release was updated to use the same major version we use for our releases. Could you check if updating it now makes it use Electron 39 and the issue is fixed?

j053ph6:
I updated Horizon but it still uses the same electron version 

<img width="1114" height="553" alt="Image" src="https://github.com/user-attachments/assets/1a4c0673-1f45-40e4-b228-c6c95cf17df2" />

FatCatClient:
If this issue doesn't happen in the AppImage or tarball releases, and it specifically started when pacman updated your system's Electron to 42-- which isn't the version we're targeting for Horizon, then it's really an issue with that third party package and we can't help here.

[Though looking at the AUR's website](https://aur.archlinux.org/packages/fchat-horizon-bin), there is a comment suggesting that you should do a clean build specifically.

[...]

PR #832, opened 2026-06-17
chore: Update electron, update pnpm, update node to match electron.

CodingWithAnxiety:
Title describes PR.

The necessary migrations have been made for PNPM 11, as described in: https://pnpm.io/migration


FatCatClient:
Let’s do this for 2.4 instead. #821 is very likely an electron 42 issue and I want to isolate those issues better.

CodingWithAnxiety:
> Let’s do this for 2.4 instead. #821 is very likely an electron 42 issue and I want to isolate those issues better.

Hey! While I agree, I do want to state on #821 I did attempt to repro it on electron 42 (Hence why I was updating, anyways) and I had no issues.

The issue may be that the electron executable they were using was incompatible with something on our end, but on that note, thanks for catching the nix flake I didn't

FatCatClient:
> > Let’s do this for 2.4 instead. #821 is very likely an electron 42 issue and I want to isolate those issues better.
> 
> Hey! While I agree, I do want to state on #821 I did attempt to repro it on electron 42 (Hence why I was updating, anyways) and I had no issues.
> 
> The issue may be that the electron executable they were using was incompatible with something on our end[...]

That may be so, but it's definitely not an isolated incident, since it's essentially a duplicate of #753. And in that specific instance, getting it to run on the supported version of Electron fixed it. It's something I want to keep an eye out for first.

And even if it's some freak incident with the AUR (many such cases!), Electron major version upgrades have definitely introduced some breaking changes, including ones that crash the app under normal usage. So I think being a bit conservative with this is indeed the play here.

CodingWithAnxiety:
Oh, no, I totally agree with waiting on 2.4, I'm just letting you know that I think the bug might be something else and that we should still investigate that possibility

FatCatClient (approved):
Let's merge this once we're finishing up the 2.3.X patch rounds and start work on 2.4.0. Just have to resolve the basic CHANGELOG.md merge conflict then.

PR #833, opened 2026-06-22
feat: Push ignored and hidden users to the bottom of the channel member list

### LINKED ISSUE #150: Feature Request: "Not Interested"
**The Problem**: Looking through profiles can be a bloated experience of having to skip people you know 100% you are not interested in.

**The Solution**: A new right click option to mark a profile as "Not Interested" This would not block, or do anything negative to the user, it would simply put them at the bottom of the userlist in a separate section for all users marked as that, for the types of accounts you don't want to roleplay with.

This would also prevent them from appearing in any character searches.

This is much friendlier way than blocking someone, because I might be fine publicly socializing with lots of people, but it would help me keep tabs on who I want to message and who  I haven't seen before.

XavierRaine:
Issue #150 discusses establishing feature parity between SlimCat's "Not Interested" and Horizon's "Hide Ads" functionality. Part of what is discussed in that issue is the prospect of resorting the full list of members in a channel to place "hidden" users at the bottom - as SlimCat does.

This PR augments the current default sort methodology - the one that places operators and friends at the top of the list - to also place hidden _and ignored_ users at the bottom of the list.

One hypothetical discussed in #150 was dimming hidden users similar to ignored users. To accomplish this, I have refactored the manner in which "isHidden" is consumed to be a property of characters similar to "isIgnored". I have also added a new hidden weight to the character profiler of -0.5.

As written, this will place all hidden users below "normal" users, and all ignored users below hidden users within the channel member list. However, ignoring or hiding a user does not cause them to be resorted within the list. Some other event must cause the list sort order to be updated before the hidden or ignored user is pushed to the bottom. Similarly, unignored or unhidden users will not return to their "normal" sort status until another sort event is triggered.

Some visual behavior within SlimCat already results in "eventual concurrence"; for example, bookmarking or unbookmarking a user causes their sort order to change, but does not update their color until the tab is reselected. I would provide the behavior to rerender the component myself, but I confess I'm not familiar enough with Vue to feel confident in an idiomatic implementation. 

Closes #150

### DIFF
--- a/chat/CharacterSearch.vue
+++ b/chat/CharacterSearch.vue
@@ -362,9 +362,7 @@
       core.connection.onMessage('FKS', async data => {
         const results = data.characters
           .map(x => ({ character: core.characters.get(x), profile: null }))
-          .filter(
-            x => !core.isHidden(x.character.name) && !x.character.isIgnored
-          )
+          .filter(x => !x.character.isHidden && !x.character.isIgnored)
           .filter(
             x =>
               this.isSpeciesMatch(x) &&
--- a/chat/UserList.vue
+++ b/chat/UserList.vue
@@ -89,7 +89,9 @@
           v-for="member in filteredMembers"
           :key="member.character.name"
           class="userlist-item"
-          :class="{ dimmed: member.character.isIgnored }"
+          :class="{
+            dimmed: member.character.isIgnored || member.character.isHidden
+          }"
         >
           <user
             :character="member.character"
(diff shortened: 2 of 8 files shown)


CodingWithAnxiety:
I want to do my own implementation of the "not interested" feature. I think I'll give it some time before I merge this, as I might have some feedback or want to commit to it myself

FatCatClient:
@CodingWithAnxiety @XavierRaine I had a few ideas for such a feature myself as well, and intended for that to sunset the 'Hide ads' feature. Something that wouldn't be a site-wide block because you might still want to talk to this person, but would save you the effort of having to find them in search results or get positive matches with them.

So yeah, let's look at this later when we're assembling 2.4 stuff. This is a good first step and I think we can add more on top of it.

FatCatClient:
This is still listed under 2.4, but I think it's better to put it on the backburner for a later release so we can properly figure out how we want this to work with a 'Not interested' toggle.

Issue #781, opened 2026-05-24
Sometimes cache events still have race conditions


FatCatClient:
### Describe the bug

Some time ago, #111 was fixed, which solves the most blatant case where this issue can happen: when you move the mouse quickly enough to have the character preview load the wrong character's data by moving to the next one before the first one has finished loading.

But there _is_ a second case where this can happen, though it's a lot more niche and it requires the data for the character preview to not have finished loading before opening a character profile for a different character.

### Version Information

Version: 2.2.0-beta.1
Commit: 0e12a36
Electron: 40.10.0
Chromium: 144.0.7559.236
Node.js: 24.15.0
Platform: Windows 64-bit (10.0.22631)

### To reproduce

1. Have a character whose profile data is saved to disk, but is not in active memory at the moment (easiest example to get is the character you are signed in as, but before you have manually opened their profile)
2. Come across a character you do not have loaded yet, (and possibly isn't saved to disk even)
3. Move mouse over that character's UserView component
4. Quickly open the profile from step one
5. Now the profile viewer will show incorrect data for character number 2, instead of character number 1.

### Additional context

I haven't checked if this fucks with the actual stored cache either. Will need more research.

Issue #758, opened 2026-05-16
Third Profile always struggles to connect


WinMister332:
### Describe the bug

No matter whether I'm on my home internet, or my phone's hotspot, both of which are decent internet, the first and second profiles can connect easily and right away, but no matter which profile I try to log in to for the third connection it fails about 90% of the time, always claiming "invalid ticket", I'm not sure if this is a problem with F-List or Horizon, but it's fairly consistent and it's ALWAYS the third profile.

### Version Information

Version: 2.1.4
Commit: 67331c27
Electron: 39.8.3
Chromium: 142.0.7444.265
Node.js: 22.22.1
Platform: Windows 64-bit (10.0.26200)

### To reproduce

Not sure how to reproduce.

### Additional context

As far as I know, I haven't had this issue with either Horizon's predecessor Rising, nor the official F-Chat client. it seems to be an issue as far as I can tell, with how the client communicates with the F-Chat servers. Note that I'm also not even sure this is a bug or even a Horizion issue, it could just be a F-List or local issue. But I wanted to note it just in-case because I didn't get it with the previous client, or the official client.

PR #838, opened 2026-06-25
Add/re-implement autoFetchGroupChats

snowsune:
Inspired by my sis Leanne!

This functionality looked like it existed before/was commented out. So I totally understand if there are good and real reasons to not include such a feature. autoFetchGroupChats is disabled by default here, like with received ads in channels, it will add a profile to the cache queue automatically when you have this option turned on.

### DIFF
--- a/chat/SettingsView.vue
+++ b/chat/SettingsView.vue
@@ -840,6 +840,20 @@
           ></settings-checkbox>
         </div>
       </div>
+
+      <div class="mb-3">
+        <div class="d-flex p-2 justify-content-between align-items-start">
+          <div class="w-50">
+            <label class="control-label" for="risingAutoFetchGroupChatProfiles">
+              {{ l('settings.profile.autoFetchGroupChats') }}
+            </label>
+          </div>
+          <settings-checkbox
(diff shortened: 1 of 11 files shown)


FatCatClient:
Thanks for the contribution, always appreciate it when someone tries to think about stuff like this as well.

Unfortunately we don't want this functionality back at the moment, because it's proven to be unstable with performance and only worsens problems like #781 and #758. It's why as is, extra profile data like custom colors and profile pics are only fetched from the disk and summarily discarded from the in-memory cache: it keeps them from hogging the F-List API too much and keeps memory usage down tremendously.

If we're bringing something like this back, it would likely be as part of a rewrite of the old Rising smart filter system (which would actually benefit from this far more than custom name colors and pfps ever will). That system is definitely in need of a complete rewrite to actually work properly, but that would be quite the time investment and is not something we have given the priority it perhaps deserves.

Maybe we can find someone who's interested in giving it the love it needs in the future though.

snowsune:
Perhaps I'll start on that then! I totally figured *because* it was explicitly not being included that there was a reason. I appreciate you taking a peek and letting me know why!

PR #839, opened 2026-06-26
Docs: Devops flow, localization system, and maybe more

### LINKED ISSUE #837: Write up issue and PR flow
I’m just making this an issue on here so I can remember to do it. 

We want to formalise this and elaborate on the intended shared tagging language, so this should likely be added to the CONTRIBUTING document.

FatCatClient:
Closes #837

Please look over this too @CodingWithAnxiety and give your thoughts, since you have to work with this as well.

Writing documentation on the localization system will be part of this too, even if just so that we can just directly point to that for contributors.

TODO:

- [x] Write about how we will be using the project board(s) for releases (@FatCatClient)
- [x] Localization system docs (@CodingWithAnxiety)
- [ ] (Maybe?) split off the build instructions into their own document (Either one of us, if we feel like it)


CodingWithAnxiety:
I like it thus far. I'll help right the documentation on localization system since I manage that when I get home. I do have some opinions, which I will leave in review comments

CONTRIBUTING.md
```diff
@@ -100,7 +103,82 @@ Look at the `mobile` directory for more info. For `android`, we recommend you us
 
 I've completely deleted the `ios` directory, and have no intention to support ios at this time. However, for those interested, the `ios` directory was removed in [this commit](https://github.com/Fchat-Horizon/Horizon/commit/41261d1ba7043eb7dfd5a1a6331dc604ff338814), and you're more then welcome to restore it.
 
-### Project layout
+### Project layout and development flow
```
CodingWithAnxiety: I think the project layout and development flow should be two *distinct* topics, because:

> The information here isn't exactly relevant for if you want to fork Horizon or just build the project for yourself [...]

Either we need to split this file for contributors / modders + builders, or make a clear boundary between the two for the sake of simplicity.

CONTRIBUTING.md
```diff
@@ -100,7 +103,82 @@ Look at the `mobile` directory for more info. For `android`, we recommend you us
...
+
+The information here isn't exactly relevant for if you want to fork Horizon or just build the project for yourself, but it might be useful for contributors so that they
```
CodingWithAnxiety: > [...] but it might be useful for contributors so that they [sic]

So that they.. what?

CONTRIBUTING.md
```diff
@@ -100,7 +103,82 @@ Look at the `mobile` directory for more info. For `android`, we recommend you us
...
+
+#### Issue tracking
+
+Issues with the type "Bug" or "Task" are supposed to be sorted by level of priority, as tagged on the GitHub issue tracker. Though the levels are both self-explanatory, while also needing to be assigned loosely at the discretion of the team member triaging, we should aim to adhere to the following priorities:
+
+- **Critical**  
+  App crashes under _regular_ use, security issues, or similar problems. Likely requires a hotfix that might bypass the regular release flow and goes straight to the `main` branch. If an issue like this is found, it means we should likely drop whatever else we are working on until it's resolved. They should also be added to the appropriate milestone.
```
CodingWithAnxiety: Maybe we should define regular use? We likely should also document how merging into main works, since that's not a typical workflow.

CONTRIBUTING.md
```diff
@@ -100,7 +103,82 @@ Look at the `mobile` directory for more info. For `android`, we recommend you us
...
+- **High**  
+  App crashes under irregular circumstances, important features are broken or regressed hard. Should be resolved quickly, but does not require dropping anything else we might be working on. If (realistically) possible, these should be resolved before the next stable release. That also means they should be added to the appropriate milestone.
```
CodingWithAnxiety: Once more, "irregular circumstances" is subjective. How is this defined?

CodingWithAnxiety (changes requested):

CONTRIBUTING.md
```diff
@@ -100,7 +106,82 @@ Look at the `mobile` directory for more info. For `android`, we recommend you us
...
+
+- **Medium**  
+  Issues with obscure features, or issues that can be worked around. Issues that originate from 3.0, F-Chat Rising, or significantly older versions of Horizon mostly fall under this umbrella, unless they're also critical. Use sound judgement here. Can be added to the current, upcoming release, milestone- but depending on the existing workload or potential deadline, it might be best to just throw it under a future, patch-release milestone.
+
+  Most styling issues probably fall underneath this category, unless they seriously impede regular use.
+
+- **Low**  
+  Minor annoyances and quirks, doubly so if they're older ones. They would be good to resolve, and if the issue itself can be done quickly then we don't need to hold back on them. Low priority issues should simply not get in the way of more important ones.
+
+Issues that have not been triaged with a priority yet should be tagged with the ["Needs Triage"](https://github.com/Fchat-Horizon/Horizon/issues?q=state%3Aopen%20label%3A%22Needs%20Triage%22) tag, which should be added automatically to any new issues created.
```
CodingWithAnxiety: This statement is mostly true, but right now only `bug.yml` actually applies it. The feature-request, question, styling, etc. templates don't. We should either reword this for clarity (specifying it only applies to bug reports), or add it to all of these issue templates.

CONTRIBUTING.md
```diff
@@ -100,7 +106,82 @@ Look at the `mobile` directory for more info. For `android`, we recommend you us
...
+We track our project's progress using [GitHub's project boards](https://github.com/orgs/Fchat-Horizon/projects/2/). The priorities given in the [Issue Tracking](#issue-tracking) header are set with the 'Priority' field from this project board, not the one for individual issues. We use the following state columns:
+
+- **Backlog**  
+  The default state. A priority has not been given yet.
+- **Triage**  
+  Issues that have been given a priority-- or at least been looked at, but have not been slotted for a specific upcoming release yet.
+- **Blocks Release**  
+  For when an issue has been added to a release, but it's also _required_ to be resolved before the release can go live. Usually that means they're high priority. Issues tagged with 'Pre-release build' are given this tag, unless they're listed as "Low" priority.
+- **Does Not Block**  
+  Slotted for a release, but work has not been started yet.
```
CodingWithAnxiety, line 145: This is partly my bad since I believe I wrote / described it at a point in time, but the issue is this describes progress, which overlaps with "in progress" right now. We should reword this to clarify whether it blocks release or not, not whether or not it is in progress.

CONTRIBUTING.md
```diff
@@ -100,7 +106,82 @@ Look at the `mobile` directory for more info. For `android`, we recommend you us
...
+- **In progress**  
+  Work has been started, but it's not yet ready for review or merging. If a pull request is made for an existing issue, it's automatically assigned this state.
+- **Follow Up Before Release**  
+  Work is hopefully finished, but maintainers need to look at the pull request still. If a pull request is set to close one or more issues, and then gets assigned this state, its issues will also be set to this.
+- **~~Ready to Ship~~**  
+  _To get rid of? We're better off letting the integration automatically move them to Done and subsequently archive them._
```
CodingWithAnxiety: While I understand fully this is a placeholder, this **must** be changed prior to us making these docs public facing. We're doubtless, omnipotent, gods. We have no doubt, we question nothing. 

... In all seriousness, we either need to remove this or leave it as it's described. Discuss in PR.

CONTRIBUTING.md
FatCatClient: I can apply it to the styling one too. You cant really apply it to things like questions.
And in my opinion, feature requests are automatically lower priority than any bugs, so why give them priorities at all?

CONTRIBUTING.md
FatCatClient: I thought I had a bit there about how it's specifically impossible to define this objectively, and that you should use common sense instead. I still feel that's true.

CONTRIBUTING.md
FatCatClient: The placeholder was there because I didn't want to just remove the column without at least asking you if it was a good idea. I don't think the column has any purpose at all.

CONTRIBUTING.md
FatCatClient: It doesn't really overlap with "in progress", because it doesn't actually imply work has started.

I think a better solution would be to merge these two into a single "Planned" column and just use a 'Blocks release' tag for the issue itself. You can configure boards to show tags on individual items so that information won't be lost, especially if you give it an obnoxious red color.

CONTRIBUTING.md
FatCatClient: See above. Has to be done on feeling.

CONTRIBUTING.md
FatCatClient: Naming the section this way is probably not doing it any favours. "Project layout" also kind of implies that we're describing the directory structure or class structure, neither of which we are doing here.

I'm not against moving build + packaging information to their own file-- since I suggested it first, but this section needs a different name regardless.

Suggestions? I'm leaning towards just "Workflow".

CONTRIBUTING.md
FatCatClient: Fugg I forgot. Let me finish this.

CONTRIBUTING.md
CodingWithAnxiety: I agree on workflow, I think that's good.

CONTRIBUTING.md
CodingWithAnxiety: I didn't see that bit, and it is fairly true, but I do think a proper definition can't hurt: How many users (And some developers...) have common sense?

CONTRIBUTING.md
CodingWithAnxiety: I think we should reword it, then. For the sake of clarity.

CONTRIBUTING.md
CodingWithAnxiety: That may be worth it in the long run, can't hurt to have a few less columns, no?

CONTRIBUTING.md
CodingWithAnxiety: Let's remove it, then. Send it.

PR #840, opened 2026-06-27
feat: Highlighted search results in logs

freenutsxd:
image

<sup>it's pretty cool</sup>

Adds a highlight around every occurrence of a searched word whenever you filter logs, because I'm tired of having to read entire multi-para posts just to see where I said a specific word. It's enabled by default and works across themes, but can be disabled by deselecting the highlight icon in the input field. 

Something to note is that bbcode tags (url, color, etc.) can still be searched, but they won't be highlighted. Only the actual visible text within the tags will be highlighted, if applicable.

Additionally includes a tiny fix that prevents a whitespace-only (e.g. just spaces) input from searching, because it looks silly with highlights enabled (and is probably bad for performance).


freenutsxd:
Expanded the highlight functionality a bit more so that it works in the chat view as well. The chat view's search also ignores any white-space only searches as well, just for parity

search

scss/_chat.scss
```diff
@@ -417,6 +417,18 @@ $bookmark-color: #66cc33 !default;
   color: $text-muted;
 }
 
+// search-term highlighting, shared by the log viewer and in-chat find
+.message-search-highlight {
```
FatCatClient, line 421: Since this is part of the built SCSS, I think it's better that highlighted text directly gets its `color` value set from the `color-contrast(...)` function. We can probably live without highlighted text having the exact same color as its BBCode tag might have given it, and manually setting contrasting text that mildly matches the bbcode color for all of these is not worth the time investment.

FatCatClient (changes requested):

scss/_chat.scss
freenutsxd: `color-contrast($body-bg)` seems to add better contrast across themes and lets me cut out the system message-specific coloring, so thanks for the suggestion. Lmk any other style critiques you have

Previous
mspaint_Hhny2McjGI

New
image

PR #848, opened 2026-06-30
fix: prune unjoinable channels from groups after startup join

### LINKED ISSUE #864: Deleted channels/ channels you are banned from persist in the pinned list
### Describe the bug

Image

So you end up with this stuff.

### Version Information

...

### To reproduce

1. Pin channel
2. Get banned from it/ have the channel be deleted
3. Reconnect

### Additional context

I only really added this issue to the issue tracker so that nobody else needs to report it. There's already a PR that fixes it.

Kannamoris:
## What

Reworks how pinned/grouped channels that can no longer be joined (e.g. deleted server-side) are cleaned up. Horizon currently re-attempts to join removed channels on **every** startup. FServ never reports which joins failed, and the incidental cleanup that used to mask this disappeared when channel grouping landed. This reimplements that cleanup as a deliberate step.

## Background

The old auto-removal was never an intentional feature; it was a side-effect of how `savePinned()` worked. It rebuilt the pinned list from the channels that actually had live conversations:

```ts
this.pinned.channels = this.channelConversations.filter(x => x.isPinned).map(x => x.channel.id);
```

A removed channel never opens a conversation, so it quietly dropped out the next time the list was saved. It worked, but only by accident. Channel grouping changed `savePinned()` to derive the list from persisted group membership instead, which removed that side-effect, and with it the cleanup. Dead channels now sit in a group forever and keep getting re-joined.

## How it works

Rather than leaning on a side-effect again, the cleanup is now explicit. A short "settle" timer is armed after the startup auto-join and **restarted on every successful join**. When joins stop arriving, any grouped channel that never produced a live conversation is treated as removed: it's pruned from its group (via the existing `syncGroupChannels` + `saveChannelGroups` + `savePinned` path) and the user is shown a console notice listing the dropped channels.

## Slow connections

The timer measures *silence between joins*, not time since login, so it can't fire while channels are still trickling in. On a slow link the cleanup just happens later, never mid-join. It also disarms on disconnect/reconnect and no-ops if the socket is closed, so a flaky connection cancels the cleanup rather than misfiring. The window is a single tunable constant (`PINNED_CLEANUP_SETTLE_IN_MS`, 10s).

## Testing

- Pinned/grouped a mix of valid channels and one removed server-side; confirmed the valid ones join, and ~10s after the last join the removed one is pruned from its group with a console notice, and is no longer attempted on restart.
Fixes #864

### DIFF
--- a/chat/conversations.ts
+++ b/chat/conversations.ts
@@ -32,6 +32,15 @@ import isChannel = Interfaces.isChannel;
  */
 const CONVERSATION_CACHE_UPDATE_FREQ_IN_MS = 1000;
 
+/**
+ * @constant
+ * How long the startup auto-join must be quiet (no new channel join arriving)
+ * before pinned channels that never joined are treated as removed and pruned.
+ * The timer resets on every successful join, so a slow connection that delivers
+ * joins gradually keeps deferring the cleanup until the joins actually stop.
+ */
+const PINNED_CLEANUP_SETTLE_IN_MS = 10000;
+
 function createMessage(
   this: any,
   type: MessageType,
@@ -782,6 +791,8 @@ class State implements Interfaces.State {
   settings!: { [key: string]: Interfaces.Settings };
   modes!: { [key: string]: Channel.Mode | undefined };
   channelGroups: Interfaces.ChannelGroup[] = [];
+  pinnedCleanupTimer?: ReturnType<typeof setTimeout>;
+  pinnedCleanupArmed = false;
 
   get channelGroupAssignments(): { [channelId: string]: string } {
     const map: { [id: string]: string } = emptyMap();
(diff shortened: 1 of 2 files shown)


chat/locales/en_us.json
```diff
@@ -142,6 +142,7 @@
   "channel.group.newGroup.counter": "New Group ({0})",
   "channel.group.noticePinned": "[b]Your pinned channels have gotten an upgrade, and now you can categorize them in groups.\n\nClick the + button in the channel list to create a new group. You can also drag your channels around, or right click channels and groups if you have any. Go ahead and give it a try!\n\nCheck out [url=https://horizn.moe/docs/channel-groups.html]our website[/url] for more info.[/b]",
   "channel.group.pinned": "Pinned",
+  "channel.group.removedUnjoinable": "[b]The following channels could no longer be joined (they were likely deleted) and have been removed from your groups:[/b] {0}",
```
FatCatClient:
```suggestion
  "channel.group.removedUnjoinable": "[b]The following channels could no longer be joined, as they were likely deleted, and have been removed from your groups:[/b] {0}",
```

chat/conversations.ts
```diff
@@ -847,6 +858,47 @@ class State implements Interfaces.State {
...
+    if (removed.length === 0) return;
+    this.syncGroupChannels();
+    void this.saveChannelGroups();
+    void this.savePinned();
+    const names = removed.map(id => {
+      const recent = this.recentChannels.find(c => c.channel === id);
+      return recent !== undefined ? recent.name : id;
+    });
+    void this.consoleTab.addMessage(
+      new EventMessage(l('channel.group.removedUnjoinable', names.join(', ')))
```
FatCatClient: Maybe these names can be wrapped in the appropriate channel BBCode. That way a user can easily click the channel names and see if the channel still exists (or if they were banned from it).

Kannamoris: Done in 2b4421d9. Private rooms are wrapped in `[session=name]id[/session]` and public channels in `[channel]name[/channel]`, same logic as the "Copy channel code" action in ChannelMenu, so clicking a name attempts the join and shows whether the channel is gone or you're banned.

PR #856, opened 2026-07-03
Fix Image Preview Stuck Open

min-xy:
This is a strange one. Occasionally, an image preview can get 'stuck' open. In some cases, the link that the preview came from is under the preview, and requires moving the mouse around until the user can luckily mouse over it again to have it close.

To resolve this, I've removed the hasMouseMovedSince check that would promptly return and cancel the dismiss of the preview.

I believe the culprit behind this bug is that when a link is in chat or the console, a new message can cause the link itself to jump, without the mouse being able to do. Timing that up results in the image not being dismissed, but staying stuck open until the user can find the link behind the preview.

I have been running these changes locally, and I have not noticed any issues. But the bug itself is a pain to reproduce, and even in normal use I would only see it once or twice a week at most.

I also removed the force bool from the function and its calls, as it was only used for that line.


FatCatClient:
Thanks for the bug fix. I've been meaning to make a bug report for it myself, but haven't gotten around to that yet.

PR #871, opened 2026-07-06
feat: device sync

Kannamoris:
## What

Adds a Device Sync section to the Data Manager that lets a Solstice device merge chat logs with Horizon over the LAN, no cable or cloud account required. Horizon shows a QR code, Solstice scans it, and both devices end up with the union of all messages. Only chat logs move; settings, character settings, drafts, pins, recents and hidden lists are never touched.

## Background

Chat logs currently live only on whichever device wrote them. Getting a phone's logs onto the desktop (or vice versa) meant manually exporting and importing a zip, which merges at the file level: a conversation either gets skipped or fully overwritten. That's fine for a one-off backup, but it's the wrong tool for keeping two actively-used devices in sync, since you lose whichever side you don't pick.

This adds a purpose-built path for that: a message-level merge that keeps both devices' messages instead of picking one file over the other.

## How it works

Horizon acts as the HTTP server for a single-use session; Solstice is the client and drives the exchange:

```
scan QR -> verify account matches locally -> POST /v1/handshake
        -> GET /v1/logs   (merge into local store)
        -> POST /v1/logs  (desktop merges)
        -> POST /v1/finish
```

The QR code carries the desktop's LAN addresses, an ephemeral port, a random session token, a random AES-256-GCM key and the account name (so Solstice can bail out with a friendly error before attempting a handshake if it's signed into the wrong account). Those secrets never cross the network any other way: every request carries the token as a bearer header, checked in constant time, and every non-empty body is AES-256-GCM encrypted. A LAN sniffer that replays the token still can't read a payload, forge a handshake, or tamper with a response; the worst it can do is disrupt the session, which the user resolves by starting a fresh one. Sessions are single-use, expire after 10 minutes without a successful handshake, and abort after 5 failed auth/decryption attempts.

The transferred payload reuses the existing export zip layout (JSON logs per character), plus a `logs-names.json` so a channel's display title survives to a device that has never seen that channel before.

The new merge (`electron/services/sync/log-merge.ts`) is the part that actually differs from the zip importer. Per conversation, it dedupes incoming messages against the existing binary log on the exact `(time, type, sender, text)` tuple, inserts whatever's left sorted by time (stable, so same-timestamp messages keep each device's original order), and rebuilds the `.idx` day index in the same pass. If nothing is new the file isn't rewritten. It's idempotent by construction, so scanning the same QR code twice adds nothing the second time.

Sync is only available while signed out, matching the existing zip importer's constraint, and a running session aborts if a character connects mid-transfer, since the chat renderer's append-only log writer and in-memory day index aren't safe to race. Solstice mirrors this on its side: Device Sync is only reachable from the character-select screen, before any character has connected.

The full wire protocol is written up in `docs/log-sync-protocol.md`. Solstice's iOS and Android apps already implement it as the client, and the two sides have been verified to interoperate byte-for-byte: same zip layout, same AES-256-GCM framing, same merge semantics. On mobile the WebView can't make these requests itself (plain HTTP to a private LAN address is blocked by ATS/local-network privacy on iOS and by cleartext policy on Android, and a `file://` origin fetch would need a CORS preflight Horizon never answers), so each platform routes through a small native bridge (`NativeSync.swift`, `Sync.kt`) that is a dumb byte transport; all crypto, zip handling and merging stay in JS on both ends. `scripts/sync-client-sim.mjs` remains useful for exercising the server without a phone.

## Testing

- Ran a full sync against real Solstice builds on both iOS and Android, over Wi-Fi: handshake, pull, push, finish, with overlapping and disjoint conversations on each side, confirming both devices converge on the same message set.
- Ran a full sync between two local data directories via `scripts/sync-client-sim.mjs`: handshake, pull, push, finish, with overlapping and disjoint conversations on each side.
- Verified re-running the same sync (same and regenerated QR, from either the desktop or a phone) adds zero messages the second time.
- Verified wrong-account, expired-session and bad-token cases each fail with the documented error and leave existing logs untouched.
- Verified a character connecting mid-session aborts the sync.


CodingWithAnxiety:
Hello!

I'll be reviewing this, today.

FatCatClient:
Does the process seemingly get stuck if the client dies prematurely? I killed the process for the client right when the server connected to it, and I'm not sure it takes that into account at all.

I don't have an actual Solstice install to test this on- so I'm just using the script for now, but there might be a couple of instances where we can expect the client to die or otherwise disconnect that this ought to take into account.

Kannamoris:
Been a little while since I worked on this, and I just checked it. I did add an expiryTimer but it only gets fired when the state is 'waiting'. So the moment the handshake takes place it gets cleared and just never arms again. I'll just re-arm the timer in every non-terminal state. Think 2 minutes per post-handshake state is enough?

CodingWithAnxiety:
Okay, so I did a full review of this PR, and have some thoughts.

In short: The PR is sane, and I generally like this feature. I think the protocol itself is pretty well thought out, too. I did, however, find two things I consider blocking, and then a handful of smaller things I'd like fixed, or at least looked at and documented, before I consider merging this.

### A character can connect while logs are being merged

This isn't necessarily a problem every time, but it is a race, and the failure mode is losing logs. So... I think it does need to be dealt with.

Right now, the signed-out check and the call to abort both live in `sync-ui.ts` / `Exporter.vue`, while `mergeLogsZip()` blocks the exporter renderer. The main process can send `connected-characters-updated`, but the exporter won't actually see it until the merge is finished. The chat window doesn't have that problem, because it's a separate renderer. It can connect and start writing messages in the meantime.

So, if the merge has already read a log, and chat appends a message to that log before the merge renames its replacement over it, that message isn't in the replacement, and thus gets overwritten.

There's also a smaller window for the same thing during startup. `startSyncSession()` checks for connected characters, waits for `LogSyncServer.start()`, and only assigns `activeServer` afterwards. If a character connects during that wait, the watcher tries to abort while `activeServer` is still undefined, and the server can finish starting anyway.

I think the actual lock has to live in the main process. It needs to know that a sync is happening and refuse a new connection until it's done. I'm not particularly attached to _how_ this is done, but I don't think the renderer watcher can be the thing protecting the files, it isn't reliable.

### The log is replaced before the new index is known to be valid

This one is less theoretical; I reproduced it locally.

Right now, `mergeLogFile()` does this:

```ts
fs.writeFileSync(tempFile, Buffer.concat(chunks));
fs.renameSync(tempFile, file);
fs.writeFileSync(`${file}.idx`, buildIndexBuffer(name, indexEntries));
```

I found at least two inputs which pass validation and then make `buildIndexBuffer()` throw:

- A message with `time: 0` is considered valid. In US/Eastern, `dayNumber(0)` is `-1`, which `writeUInt16LE()` rejects.
- `logs-names.json` can contain a name longer than 255 UTF-8 bytes, which `writeUInt8()` rejects.

These are admittedly weird inputs. `time: 0` fits the binary log format, while the oversized name would have to come from a malformed or third-party archive. Still, the importer accepts both. In both cases, the replacement log was already on disk by the time index creation failed, and no new index was written.

Retrying doesn't repair it, either. The message is now in the log, so it gets deduplicated, `added.length === 0`, and the function returns without rebuilding anything. If the conversation already had an index, that old index may now have offsets for the file from before it was sorted.

Could we build the index before replacing the log? We already have `buildLogIndexBuffer()` in `electron/services/log-backup.ts`, and it handles both the name truncation and the day range. I'd rather reuse that if possible than have two index builders with different validation rules.

### The smaller stuff

I don't think any of these need their own sections, but I do want to ensure they're not lost.

- If `GET /v1/logs` throws while preparing the response, the server stays in `sending`, and the expiry timer has already been cleared. Every retry after that gets a 409. The upload path resets its state after an error; I think the download path needs to do that as well.

- It looks like a quick double-click can get two `startSyncSession()` calls past the initial `activeServer` check before either one has assigned it. Also, because `activeServer` belongs to the renderer, another Data Manager window gets another copy of that guard.

- The 512 MiB limit only applies to the encrypted upload. The zip can expand well beyond that in memory, and `AdmZip` doesn't appear to have an uncompressed-size limit here. The download side has no equivalent limit, either, and briefly keeps several full copies of the archive in memory. I'm not sure what the exact limits should be, but I do think both directions need one.

- I tried putting a `Thumbs.db` in the log directory, and it was included in both the export and the conversation count. I had this issue before and put together a   `isFilesystemArtifact()` helper, so this should hopefully just be a matter of using it here too.

- `isValidLogMessage()` accepts any message type from 0 through 255, while the enum and protocol only define 0 through 6. I think it should check the range we actually support to avoid weird edge cases.

- One protocol/docs mismatch: the documented `busy` response can't currently happen. The state check runs first, so a second request gets `wrong_state` instead. Either the order or the documentation should change.

Everything else passes: Typechecking, I18n, Prettier. Nothing additional to note.

FatCatClient:
I’ll quickly add my 2c on the upload limit: My log backup(s) hover around 275mb right now. I don’t find it unreasonable to believe that there are people who have larger log collections than 512mb. 

Should we accommodate those power users too, and if so, how?

Kannamoris:
@FatCatClient @CodingWithAnxiety What uncompressed-archive cap should I enforce when merging? Does 2 GiB sound good?

CodingWithAnxiety:
> @FatCatClient @CodingWithAnxiety What uncompressed-archive cap should I enforce when merging? Does 2 GiB sound good? 

I'd need to sit down when I get home, but that sounds sane-ish.

electron/main.ts
```diff
@@ -1655,6 +1665,41 @@ async function onReady(): Promise<void> {
       if (autoBackupScheduler) autoBackupScheduler.runOnConnect();
     }
   );
+  // Device-sync lock. The exporter renderer acquires this before starting a
```
FatCatClient: Don't we already have logic to handle blocking connection when running the exporter/ importer?

The fact that I'm not even sure anymore is already indicative that the Main <> Renderer interaction is a spaghetti mess, but if I'm right we should probably not make that problem even worse.


If we don't, I'd rather it be made generic in a way that doesn't tie it so strictly to the log sync. But if we pick that route, we should probably sit down together to discuss how that would work first.

electron/Exporter.vue
```diff
@@ -1831,6 +1951,7 @@
         this.exportAnimatedDots = '';
       },
       close(): void {
```
FatCatClient: Do we have two `close()` functions now with their own code? Which one winds up getting called?

electron/services/sync/server.ts
```diff
@@ -0,0 +1,454 @@
...
+    this.respondJson(res, 200, {
+      ok: true,
+      deviceName: os.hostname(),
+      account: this.options.account,
+      protocolVersion: SYNC_PROTOCOL_VERSION
+    });
+  }
+
+  private async handleGetLogs(res: http.ServerResponse): Promise<void> {
+    if (this.busy) throw syncError(409, 'busy');
```
FatCatClient, line 403: Half rhetorical question/ half regular question, but are `busy` and `not-paired` states that the client might need to handle differently? Having the client check that logic based on the body is annoying as fuck (but admittedly better than always giving you 200 and needing to check the error field like F-List's API does lmao).

It might be better to put some of those values like the status codes in a const, but that's a major nitpick on my end.

chat/locales/en-US.json
```diff
@@ -1032,6 +1034,31 @@
...
+  "sync.error.expired": "The sync session expired. Start a new one and scan the fresh QR code.",
+  "sync.error.generic": "Sync failed: {0}",
+  "sync.error.lockedWhileConnected": "Device sync is disabled while signed into a character.",
+  "sync.error.timedOut": "The sync session timed out because the other device stopped responding. Start a new session to try again.",
+  "sync.manualHint": "If scanning doesn't work, paste this connection code into Solstice instead:",
+  "sync.qrAlt": "Sync session QR code",
+  "sync.scanHint": "Scan this QR code with Solstice to sync your chat logs.",
+  "sync.start": "Start sync session",
+  "sync.state.merging": "Merging received logs...",
+  "sync.state.paired": "Connected to {0}.",
```
FatCatClient: I don't think this was merged into dev when you started work on this, so you probably missed this. But language strings now use named placeholders with named fields in the function call. Helps translators a ton too, because they can now see what the placeholder is meant to represent.

```suggestion
  "sync.state.paired": "Connected to {pairedDevice}.",
```

Which you would then reference by calling it like so (for example):

```ts
l('sync.state.paired', pairedDevice: getDeviceName());
```

electron/services/sync/log-merge.ts
```diff
@@ -0,0 +1,313 @@
...
+import * as path from 'path';
+import { buildLogIndexBuffer, isFilesystemArtifact } from '../log-backup';
+import type { LogMergeStats } from './protocol';
+
+// Highest Conversation.Message.Type enum value (Bcast); see chat/interfaces.ts
+// and docs/log-sync-protocol.md. Kept as a literal because this module is pure
+// Node and must not import chat/.
+const MAX_MESSAGE_TYPE = 6;
+
+export interface LogMessage {
```
FatCatClient: Whats functionally the difference between this and ``JsonLogMessage`` in ``electron/services/log-backup.ts``?

Idem for other functions in these two files, like `parseBinaryLog()` being a match for `binaryLogToJson()` from the backup file, at least at a cursory glance. Which you are already using in the `logs-zip` file.

If there's already some code that's very similar to what you need, but it might need some specific tweaks, you might be better off making that existing function a bit more flexible rather than essentially duplicating it.

Please check this for other functions and types in this file too.

electron/services/sync/log-merge.ts
```diff
@@ -0,0 +1,313 @@
...
+  for (const entry of zip.getEntries()) total += entry.header.size;
+  return total;
+}
+
+/**
+ * Merges every `characters/{char}/logs/{key}.json` entry of a sync zip
+ * (the logs-only export format, see docs/log-sync-protocol.md) into the
+ * local log store at `dataDir`.
+ */
+export function mergeLogsZip(dataDir: string, zip: AdmZip): LogMergeStats {
```
FatCatClient: I get that this refers to the docs too for some of it, but I can't say I find this an easy to parse function. Even a few better variable names would go a long way here.

The docs should explain global concepts or interfaces for other tools, programs, or whatever to work with. Not why this function specifically needs segment 2 in the split array to be the `logs` directory name.

electron/services/sync/logs-zip.ts
```diff
@@ -0,0 +1,135 @@
...
+function listLogFiles(logsDir: string): string[] {
+  try {
+    return fs
+      .readdirSync(logsDir, { withFileTypes: true })
+      .filter(
+        entry =>
+          entry.isFile() &&
+          !entry.name.endsWith('.idx') &&
+          !entry.name.endsWith('.syncmerge') &&
+          !isFilesystemArtifact(entry.name)
```
FatCatClient, line 63: isFilesystemArtifact doesn't check for something like .DS_STORE files iirc. Which isn't your fault, but would probably need to be handled in there too and seeing it being used here reminded me to double check.

@CodingWithAnxiety

FatCatClient (commented):
The architecture of the temporary server instance seems fine to me. I have a couple of issues with how it gets handled from the UI layer straight into logic, but that's not a new issue. It's more so that I do not like the pattern that was in place, and that adding on top of it made me realize that. 

But I have similar issues elsewhere too where I feel the vue components are calling logic that should really be decoupled entirely, or UI state is handled in a really weird way. Not something for this PR, I think. The existing patterns were just messy.

For the UX I'm happy that the process itself is fairly simple. At most, I'd add a few steps on how to get the QR code scanner function in Solstice, but that's not a big deal either way. Kept it very simple to use, which would be the most important part.

chat/locales/en-US.json
Kannamoris: Yep, this predated my branch and I just missed the switch to named placeholders. Fixed.

electron/Exporter.vue
Kannamoris: Fixed: folded them into one `close()` that keeps the busy prompt and only tears down the sync session once you've actually committed to closing.

electron/services/sync/logs-zip.ts
Kannamoris: The `^\.` branch matches anything starting with a dot, so `.DS_Store` and the `._*` AppleDouble files are both caught. The docstring on the function even lists `.DS_Store` as one of the cases. So this usage is already screening it.

electron/services/sync/log-merge.ts
Kannamoris: Fixed: pulled the whole path check into a `parseLogEntryPath(entryName)` helper that destructures the split into named parts (`[top, character, kind, file]`) instead of indexing and hands back `{ character, key }` or undefined, so the loop is now just "parse the entry, skip it if it isn't a valid log file, merge."

electron/main.ts
Kannamoris: No, we don't. The import/export "in progress" flags never reach main, and the Data Manager modal isn't a real guard (blocks one window, "New Window" sidesteps it, and connect is socket-driven on auto-login/reconnect). Import already writes into the live logs dir uncoordinated, so that race exists today, sync is just the first thing to guard it. If you want I can generalize the lock into one "exclusive data session" that import/export also take, which would close that race too.

electron/services/sync/log-merge.ts
Kannamoris: Fixed: import from log-backup, keeping only the sync-specific bits (`isValidLogMessage`, the dedupe/union merge, path guards, `mergeLogsZip`). `jsonLogToBinary` even null-guards `sender` where mine didn't.

electron/services/sync/server.ts
Kannamoris: Yeah, they're handled differently: `busy` = a transfer's already running, so wait and retry the same call; `not-paired` = you hit a logs endpoint before the handshake landed, so retrying won't help, (re)handshake first. Both are 409 because both are "request conflicts with current session state," with the body `code` disambiguating, which beats always-200-check-the-error-field. If you'd rather branch on status alone I can split them onto separate statuses. And good call on the const, happy to pull every (status, code) pair into one `SYNC_ERRORS` map in protocol.ts if you want it.

FatCatClient: > And good call on the const, happy to pull every (status, code) pair into one SYNC_ERRORS map in protocol.ts if you want it.

It’s probably something I’d throw into a shared helper myself for use in other places. But I don’t think it’s terribly important either way. Just that it would make it a little bit easier to read if status codes are some kind of enum value.

electron/main.ts
FatCatClient: @CodingWithAnxiety Thoughts?

CodingWithAnxiety:
I'm going to do a final review over this, @Kannamoris. If I stumble upon anything noteworthy, you will be informed, otherwise, it will be merged.

electron/main.ts
CodingWithAnxiety: > @CodingWithAnxiety Thoughts?

Okay! So, after reading this, I think Kanna may have a point here. We *do* have a UI state around these existing ops, but it's not a "lock" per-se, it only protects that particular renderer, and another window can sidestep it.

I think we should make this generic, and use it across all our import/export ops (They must aquire the lock before writing logs!), since we're already here I think this fits in this scope. TLDR, yes to making the lock generic no, we don't need to refactor the Main Pro. <> Renderer Pro. relationship as part of doing it.

CodingWithAnxiety:
I have some concerns, but I'll address them a little later. There are a few blocking ones I didn't catch originall I'd like to address, first.

CodingWithAnxiety:
I went through your PR once more, @Kannamoris. Your fixes cleared all my previous concerns, but I found a few more issues.

**Locks.** The new lock protects sync but not the ZIP or vanilla import flows, so they can still write during a sync. This should be a generic lock shared across all three. We have discussed this prior, so this isn't really a new finding but of note.

**Merging into a damaged log.** The parser stops at the first bad record and returns everything before it. If the phone adds a new message, the merge writes that partial result over the original file and drops the rest of the log. That's data loss. Skip that conversation instead, and tell the user to run Fix Logs.

**Stop/send lifecycle.** Two bugs:
- Press Stop while Horizon builds the ZIP, and the build keeps running anyway. When it finishes, it can move the server from `stopped` back to `paired`.
- The idle timer starts as soon as the response is queued, not when the transfer finishes. A large or slow transfer can get cut off halfway through.

A few smaller issues:
- The account check itself makes sense, since it stops Solstice from syncing while signed into the wrong account. The problem is where Horizon gets its copy. It currently reads the saved-login settings, but a successful login only updates those when Save Login is enabled. On a fresh install, someone can log in successfully without saving their credentials and still get the "sign in at least once" error when starting Device Sync. This should use the most recent successful login instead.
- Message-size validation accepts records the binary serializer can't write.
- The plaintext temp ZIP is readable by other local users on Linux. Whether that's worth fixing is your call, given that logs themselves aren't encrypted. 
- Two slow handshakes can both succeed.
- The `.idx` check is case-sensitive. An incoming `general.IDX.json` passes validation, but on Windows and macOS the resulting `general.IDX` path can alias the real `general.idx` file. Sync can then treat that index as a conversation and overwrite it, leaving the original log with broken history dates/offsets or corruption warnings. Normal Solstice shouldn't produce this, but the check should still be case-insensitive.
- An index write failure leaves the new log paired with the old index. Rare, but can be catastropic.

As a notice: This PR comment was partially generated using AI, but the findings are that of my own and have been verified.

PR #882, opened 2026-07-08
fix: improve colorblind mode contrast and highlight visibility

### LINKED ISSUE #538: Colorblindness settings are awful, and I had trouble seeing the highlights
**Describe the bug**
Colorblindness settings are awful, and I had trouble seeing the highlights. Dark Green and Light Green look very close, and the colorblindness mode is all very close shades of light blue... 

**To Reproduce**

1. Go to 'settings'
2. Click on 'apperence'
3. Scroll down to 'colorblindess mode'
4. Witness the horror.

**Expected behavior**
Higher contrast, legit the 3.0 / site default colors are far better and have more contrast. Maybe use Light Blue, you can keep Dark green as place for Green, Amber is fine, Red might need a change for people with that issue. 

I have very mild R/G colorblindness at most, but still this makes it a pain to use.

**Screenshots**

- OS: Windows 10 long Term Support. 
- Version 1.35.3

**Additional context**
fatcat said I can report this as a Bug.

Kannamoris:
Fixes #538

## Problem

The colorblind mode palette (shared by all 14 chat themes) contained several pairs that are indistinguishable under red/green color vision deficiency: simulating protanopia/deuteranopia (Machado 2009 matrices) gives a CIELAB deltaE of 9 for shemale vs herm, 13 for transgender vs cunt-boy, and 16-20 for male vs male-herm (below ~20 is hard to tell apart as text color). On the four light themes the palette was also nearly unreadable (female had 1.1:1 contrast against the background). Message highlights derive from each theme's green `$success` and were not touched by colorblind mode at all, which is exactly the hue R/G-colorblind users struggle with.

## Changes

All inside the `.colorblindMode` scope in `scss/themes/_colorblind.scss`; normal mode is untouched.

- Replace the single `$genders` palette with two Okabe-Ito-derived, CVD-tuned palettes, selected per theme via `color.lightness($body-bg)`:
  - Dark themes: every pair keeps deltaE >= 17 under normal, protanopia and deuteranopia vision, with >= 4.0:1 contrast against the background. Male stays a light blue and cunt-boy stays green, as suggested in the issue.
  - Light themes (light, snowed-in, peached, no-exceptions): deltaE >= 15 and >= 3.9:1 contrast for every color.
- Restyle message highlights in colorblind mode with a blue background tint and an amber left border (classic and modern layouts), plus a blue last-read divider, so the highlight cue no longer relies on the red-green axis.
- Fix a pre-existing typo (`#{gender}` missing the `$`) that emitted a broken `.gender-icon-gender` selector for message events.

### DIFF
--- a/scss/themes/_colorblind.scss
+++ b/scss/themes/_colorblind.scss
@@ -20,15 +20,36 @@ $scoreStandoutWeakMismatchBorderColor: #888;
 $scoreStandoutWeakMismatchBgColor: rgba(170, 170, 170, 0);
 $scoreStandoutMismatchBorderColor: #ab0000;
 
-$genders: (
-  'shemale': #e2445f,
-  'herm': #d40759,
-  'none': $gray-500,
-  'female': #66ff74,
-  'male': #6699ff,
-  'male-herm': #007fff,
-  'transgender': #ee8822,
-  'cunt-boy': #63ee22
+// Palettes tuned for red/green colour vision deficiency: under protanopia and
+// deuteranopia only the blue-yellow axis and lightness survive, so hues drift
+// from the normal-mode palette to keep every pair distinguishable. Validated
+// with a Machado (2009) CVD simulation; any edit must keep a pairwise CIELAB
+// deltaE of at least 17 (dark) / 13 (light) under normal, protan and deutan
+// vision, and stay readable against the theme background.
+$cvd-light-theme: color.lightness($body-bg) > 50%;
+
+$genders: if(
+  $cvd-light-theme,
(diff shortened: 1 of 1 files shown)


Kannamoris:
light
default

FatCatClient:
I think realistically a _proper_ colorblind mode would just enable the gender icons, since I think that the best color blind accessibility is to simply not relay info based on colour at all. But also, I can't be assed to do that and will probably just move the setting for that into a global accessibility setting at some later point in time. 

Looks good to me.

PR #883, opened 2026-07-08
fix: key message drafts by conversation key instead of display name

### LINKED ISSUE #409: Channel message drafts are saved per conversation name...
**Describe the bug**
...instead of per channel ID. Which can lead to some interesting cases where if you join a different user-made channel with the same name (which is very much possible), or open a conversation with someone whose name matches that of a channel you have drafts for, you wind up crossing them.

User-made channels normally have an ID starting with ``adh-`` for this reason, so message drafts should try and use that instead of the display name.

**To Reproduce**

1. Join any channel with a name that can also be a character name
2. Write a message, but don't send it
3. Close the tab
4. Reopen the tab and sign in as the same character again
5. Open a PM with a character using this name (whether the character exists or not)
6. See your draft message appear

- OS: All
- Version 1.33.6

Kannamoris:
Drafts were cached under the lowercased conversation display name, so a user-made channel and a PM (or another channel) sharing that name would cross drafts. Conversations already carry a collision-safe key (#<channel.id> for channels, lowercased character name for PMs, _ for the console) that logs and per-conversation settings use, so drafts now use it too. Existing name-keyed channel drafts on disk are intentionally not migrated.

Fixes #409

### DIFF
--- a/chat/conversations.ts
+++ b/chat/conversations.ts
@@ -1262,6 +1262,9 @@ async function initConversationCache(this: Conversation): Promise<void> {
   // Restore message draft if it exists (e.g. accidentally closing the tab). Be sure the cache is reset for a new character if needed.
   await core.cache.conversationDraftCache.resetCacheIfNeeded();
 
+  // Only an open conversation knows both the display name pre-key drafts were stored under and its unique key.
+  core.cache.migrateConversationDraft(this.name, this.key);
+
   const draft = core.cache.getConversationDraft(this.key);
   this.enteredText = draft;
 
--- a/chat/conversations.ts
+++ b/chat/conversations.ts
@@ -137,7 +137,7 @@ abstract class Conversation implements Interfaces.Conversation {
   clearText(): void {
     setImmediate(() => {
       this.enteredText = '';
-      core.cache.conversationDraftCache.deregister(this.name);
+      core.cache.conversationDraftCache.deregister(this.key);
     });
   }
 
@@ -475,7 +475,7 @@ class PrivateConversation
       this.safeAddMessage(message);
 
       await this.logMessage(message, false);
-      core.cache.deregisterConversationDraft(this.name);
+      core.cache.deregisterConversationDraft(this.key);
       this.markRead();
     });
   }
@@ -1262,7 +1262,7 @@ async function initConversationCache(this: Conversation): Promise<void> {
   // Restore message draft if it exists (e.g. accidentally closing the tab). Be sure the cache is reset for a new character if needed.
   await core.cache.conversationDraftCache.resetCacheIfNeeded();
 
-  const draft = core.cache.getConversationDraft(this.name);
+  const draft = core.cache.getConversationDraft(this.key);
   this.enteredText = draft;
 
   if (!this.cacheActive) {
@@ -1282,8 +1282,8 @@ async function initConversationCache(this: Conversation): Promise<void> {
       }
 
       this.enteredText
-        ? core.cache.registerConversationDraft(this.name, this.enteredText)
-        : core.cache.deregisterConversationDraft(this.name);
+        ? core.cache.registerConversationDraft(this.key, this.enteredText)
+        : core.cache.deregisterConversationDraft(this.key);
     }, CONVERSATION_CACHE_UPDATE_FREQ_IN_MS);
   }
 }
--- a/learn/cache-manager.ts
+++ b/learn/cache-manager.ts
@@ -560,42 +560,42 @@ export class CacheManager {
   /**
    * Register the given draft in the draft cache.
    * @function
-   * @param {string} channel
-   * The intended recipient of the message, either a character name or a channel name.
+   * @param {string} key
+   * The unique key of the conversation the draft belongs to, e.g. `#adh-...` for a channel or a lowercased character name for a PM.
    * @param {string} message
    * The draft text as it currently exists in the input textbox.
    * @internal
    */
-  public registerConversationDraft(channel: string, message: string): void {
+  public registerConversationDraft(key: string, message: string): void {
     this.conversationDraftCache.register({
-      channel,
+      key,
       message
     });
   }
 
   /**
-   * Removes any existing draft from the cache for a given channel.
+   * Removes any existing draft from the cache for a given conversation.
    * @function
-   * @param {string} channel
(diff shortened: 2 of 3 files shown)


CodingWithAnxiety:
i just mereged my fix for this instead of yours oh my god i didnt see this im gonna die

CodingWithAnxiety:
yours is more comp[lete anyways so i reverted mine

CodingWithAnxiety:
... Okay. So, this is a long winded explanation as to why I won't merge this *yet*.

In short: The fix itself is right, yes. Every draft calls through `Conversation.key`, and so **new** channel drafts never collide with PM drafts. Great, perfect. However: 

The `channel` => `key` rename doesn't prevent old draft files from loading.  There's _no_ shape-aware deserialization anywhere in `getDrafts()`, it's just plain ol' `JSON.parse()`, and the only property ever read is `.message`. So pre-2.3.2 records load fully intact under their old lowercased name keys, which is fine for old Pm drafts. The issue is old channel drafts now sit in the same bare-name namespace PM lookups use, meaning that PMs which have a matching name will match, and thus load the now orphaned channel drafts. This means the issue in #409 is still preserved in some form, just with pre-upgrade data.

We should drop pre-2.3.2 `{channel}` style records which would fix this bug completely, but a more robust solution would be replacing `JSON.parse()` with something a little more intelligent. Maybe we could version drafts like we do exports, @FatCatClient ? Tie draft versions to the client version and when theres a breaking change we could do some kinda migration? Maybe that's over complicated tho, give your thoughts.

CodingWithAnxiety:
also, @Kannamoris , do you want me to fix this or do you want to?

FatCatClient:
> We should drop pre-2.3.2 {channel} style records which would fix this bug completely, but a more robust solution would be replacing JSON.parse() with something a little more intelligent. Maybe we could version drafts like we do exports, @FatCatClient ? 

If you're going by the book, you'd have to do this for every breaking format change, yes.

In this case, the initial import of the file would have to check if it is a pre-v2 version from the JSON and then purging incorrect data which does not exist in the defined type, before casting it to the type in question.


Fuck, this stupid JSON.parse() shit is why I can't stand Typescript sometimes. It's not even a flaw with TS's APIs, just the fact that it gets plopped right on top of regular Javascript like that.

Kannamoris:
> also, @Kannamoris , do you want me to fix this or do you want to?

I'll take care of it.

Kannamoris:
Pushed a fix. `populateCache()` now runs everything from disk through a sanitize pass that only keeps records matching the current `{key, message}` shape and re-keys them from the record itself, so the old `{channel}` records (and whatever other junk ends up in that file) just get dropped before anything can look them up. Basically what @FatCatClient described: purge anything that doesn't match the type before trusting it. And since the shape difference is unambiguous, it doubles as the version check.

One thing I did on purpose is that it filters on every load instead of migrating once. Backup import restores `drafts.txt` verbatim, so someone can reintroduce an old-format file months from now and a one-time migration would happily miss it. Filtering at the load boundary covers that forever, and the next periodic save rewrites the file clean anyway. Downside is old PM drafts get dropped too (nothing in the old records says which kind they were), but they're drafts. If you cared, you'd have hit send.

For device sync I also looked through how exports do it, and funny enough the manifest stamps `version: 2` but nothing in the importer ever reads it. The thing doing the real work there is the `isValidManifest` shape check, i.e. exactly this. So I'd hold off on a version field until the format actually grows. If we ever want one, `{version, drafts}` can be its own PR with this sanitize as the fallback for bare maps.

FatCatClient:
> For device sync I also looked through how exports do it, and funny enough the manifest stamps `version: 2` but nothing in the importer ever reads it. The thing doing the real work there is the `isValidManifest` shape check, i.e. exactly this. So I'd hold off on a version field until the format actually grows. If we ever want one, `{version, drafts}` can be its own PR with this sanitize as the fallback for bare maps.

I think what we really need is a generic shared method somewhere for parsing JSON to an object in a way that actually respects the intended type and its fields, which is another good way to prevent format issues: It can just check a version number and then fill in any missing fields it would otherwise assume it has, if necessary. No purging of orphaned fields required. 

Maybe a refactor for the future.

FatCatClient:
Anyway, after testing this for a bit, it seems to work exactly how we expected it would.

> Existing name-keyed channel drafts on disk are intentionally not migrated.

If this isn't a problem in @CodingWithAnxiety's eyes (and it's something I feel a bit iffy about, but not too much), then we can just merge it and finally cross it off.

Just as long as we properly tell people that this will happen in the changelogs, I guess.

CodingWithAnxiety:
> Anyway, after testing this for a bit, it seems to work exactly how we expected it would.
> 
> > Existing name-keyed channel drafts on disk are intentionally not migrated.
> 
> If this isn't a problem in @CodingWithAnxiety's eyes (and it's something I feel a bit iffy about, but not too much), then we can just merge it and finally cross it off.
> 
> Just as long as we properly tell people that this will happen in the changelogs, I guess.

I'm a tad skeptical, too. I'd like to see the reasoning for this intentional decision from the OP. @Kannamoris?

FatCatClient:
To be fair, I would have made the exact same decision too, since there is no way to accurately determine if an old-style name key belongs to a PM or channel. 

I just think people might not like that side effect.

Kannamoris:
Personally thought that drafts were not really that important to keep between versions. From experience I only see people using drafts as a safeguard in case the connection unexpectedly drops. For anything else I still think the drafts system just a little too unreliable. If you folk do want some sort of migration, I could look and see if I could implement something to take care of that.

Kannamoris:
Pushed the migration. Old {channel, message} records are no longer dropped; the sanitize pass now quarantines them in a side map that draft lookups never touch, and each conversation claims its own record when it opens, since that's the one moment both the display name and the collision-safe key are known. Unclaimed records are written back in their old shape until their conversation gets opened, so nothing is lost for rooms you don't visit on the first launch. The split still runs on every load, so a backup-imported old file months from now migrates the same way. Only genuinely ambiguous data is best-effort: a channel and a character sharing a name means whichever opens first gets the draft, which is the ceiling for what the old records can tell us.

CodingWithAnxiety:
After changes made, and testing, this seems sane.  Merging.

Issue #752, opened 2026-05-10
Online friends disappear from the list when unfriending via the site


KarmikCykle:
### Describe the bug

Removing a friend through the F-List website proper while Horizon is open causes all remaining listed friends who are online to vanish from the list, leaving only bookmarks visible. Closing and reopening the chat seems to fix it.

### Version Information

Horizon 2.1.4

FatCatClient:
Do you happen to remember if the character you removed was friends with the character you were signed in as, or was it explicitly a different character's friend?

KarmikCykle:
They were not, I was signed in as a brand new character with no friends added yet

KarmikCykle:
I also didn't have the setting to "hide non-character friends" turned on, though I did have "show this character's friends separately"

PR #884, opened 2026-07-09
fix: Friends appearing twice in user list

### LINKED ISSUE #581: Friends can sometimes appear twice (and not be removed from the online list properly)

freenutsxd:
Adds a few extra guards that prevents characters from appearing twice within the friends/bookmarks lists. Additionally, when someone logs off, they will now unconditionally be cleared from the friends/bookmarks lists so that no lingering duplicate entries can persist.

Fixes #581.

### DIFF
--- a/fchat/characters.ts
+++ b/fchat/characters.ts
@@ -323,7 +323,11 @@ export default function (this: void, connection: Connection): Interfaces.State {
       case 'trackadd':
         state.bookmarkList.push(data.name);
         character.isBookmarked = true;
-        if (character.status !== 'offline') state.bookmarks.push(character);
+        if (
+          character.status !== 'offline' &&
+          state.bookmarks.indexOf(character) === -1
+        )
+          state.bookmarks.push(character);
         break;
       case 'trackrem':
         state.bookmarkList.splice(state.bookmarkList.indexOf(data.name), 1);
--- a/fchat/characters.ts
+++ b/fchat/characters.ts
@@ -28,6 +28,14 @@ class Character implements Interfaces.Character {
   }
 }
 
+function removeAllFromList(list: Character[], character: Character): void {
+  let index = list.indexOf(character);
+  while (index !== -1) {
+    list.splice(index, 1);
+    index = list.indexOf(character, index);
+  }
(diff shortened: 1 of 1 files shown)


FatCatClient:
On the dev branch (and this is presumably an old ass bug from last year), there's a bug where if you enable ``core.state.settings.hideNonCharacterFriends`` you will not see characters sign in messages if they are considered character friends.

See ``chat/conversations.ts``: ``isOfInterest()`` and ``fchat/characters.ts``'s default export (the `Interfaces.State` implementation) handler for the `'connecting'` event. But I'm 99% sure the issue is just because ``character.isCharacterFriend`` isn't being set for characters as you connect, the way the ``isFriend`` field is set.

Can you also check for that issue? I probably can't reliably fix it in the dev branch if the surrounding logic is being improved like this, not until that's merged at least.

freenutsxd:
> On the dev branch (and this is presumably an old ass bug from last year), there's a bug where if you enable `core.state.settings.hideNonCharacterFriends` you will not see characters sign in messages if they are considered character friends.

Should be fixed now in 4536136.

FatCatClient:
Does this handle #752 as well? If not, no big deal. We can get to that later.

freenutsxd:
I don't think so? Their issue doesn't sound like some duplicate entry causing the bug, so my changes shouldn't solve for that

CodingWithAnxiety:
Merging after I made my changes. Open issue if anything is wrong.

PR #885, opened 2026-07-09
fix: apply the high quality portraits toggle without a restart

### LINKED ISSUE #736: The 'show high quality portraits' setting doesn't toggle consistently/ needs restarting
### Describe the bug

Various places where HQ portraits can show up don't actually update between regular and HQ portraits if you toggle the setting. Because they usually require the whole component to re-instantiate, this means that you either need to close and reopen a specific menu, or in some cases need to close the whole tab first.

### Version Information

Can't copy paste the info rn, but it's on 2.1.1 and happens on all platforms.

### To reproduce

_No response_

### Additional context

I don't have a concise list of all places where this is a problem, but from the top of my head I know it's an issue with the conversation list items, and the big icon on the top left too.

Some of these also have other settings that toggle the individual behavior (like messages), but I don't know to what extent those settings have a similar problem or might interfere with this issue.

Kannamoris:
The HQ portrait setting was only checked at profile-fetch time and its result baked into `character.overrides.avatarUrl`, so toggling it did nothing until components were recreated or the app restarted. The override is now always extracted and stored (and cleared when a profile drops its portrait tag), and the setting is checked at render time in `characterImage()` and the few places that read the override directly, so the toggle applies immediately everywhere, including the shell window's character tab.

Fixes #736


FatCatClient:
@Kannamoris, could you resolve the merge conflicts so I can test this as if it were merged into `development`?

PR #893, opened 2026-07-15
rebuild log indexes and preserve conversation names

### LINKED ISSUE #886: JSON style exports do not import correctly and lack crucial information from index files
### Describe the bug

When creating manual exports the logs get converted to a json format. When importing these, appropriate index files are not being created. 

This might also mean that the exports lack crucial information from the conversations themselves, like ad-hoc channel names, or capitalisation. 

### Version Information

Phoneposting atm. It’s presumably from 2.2.0 and onward. 

### To reproduce

1. Create a new manual export. This might not be the case for auto exports
2. Set your log directory to an empty folder, or set up a new installation on a different machine 
3. Import your .zip file from step 1
4. Settings import correctly. When looking at the log viewer, conversations appear to be added but they have no messages. 

### Additional context

_No response_

CodingWithAnxiety:
Howdy, folks!

This is the fix for #886. originally, it was just making json imports write index files, but it grew some limbs along the way. Consider it feature complete unless review turns something up.

I intend this for Friday's release. Closes #886.

## Summary

The bug: manual exports convert logs to JSON, but importing them never rebuilt the `.idx` files the log viewer needs, so conversations imported as empty ghosts. On top of that, the display names that only exist in `.idx` files (ad-hoc channel names, proper capitalization) never traveled with the export at all.

- Imported JSON logs now rebuild their `.idx` files
  - Both the GUI and CLI importers, through a new shared `electron/services/log-backup.ts` module.
  - Rebuilt indexes are byte-identical to what Fix Logs produces. This has some unpreventable quirks, but we fix what we can.
- Conversation names now travel with exports
  - New per-character `conversation-names.json` sidecar in the ZIP, read from the `.idx` headers at export time:
    ```json
    { "#adh-b2fb2f4b0a17": "Cool Channel FRFR", "some character": "Some Character" }
    ```
  - Old clients only process `logs`, `settings`, and `drafts.txt`, so they ignore the sidecar entirely, and log entries stay bare JSON arrays. Old clients can still import new exports safely.
- Name recovery for legacy exports that have no sidecar
  - Private conversations: the log file name is just the lowercased character name, and the proper capitalization survives in the message senders, so the name is recovered from the log itself.
  - Channels: recovered from the character's `settings/recentChannels` file in the same ZIP (covers the ~50 most recently joined channels, which is what that file holds. _This is the unpreventable quirk, and can't really be solved any other way._)
  - Recovered names are the lowest-priority source: sidecar > name embedded in the log > existing index > recovered > file name. Recovery can only improve on the raw file key, never override real data.
- Re-imports heal existing damage
  - If you already imported with a broken version (logs on disk, indexes missing), re-importing the same export rebuilds the missing indexes even though the log files themselves are skipped. The summary counts these properly instead of claiming "Restored data for 0 character(s)".
  
  ## Additional
- Broken CLI imports used to write JSON logs verbatim to disk as `foo.json`. Those stale artifacts are now cleaned up on re-import, but **only** when the artifact is zero-byte (Fix Logs already truncated it) or every one of its messages provably exists in the binary log next to it. If that containment check fails, the file stays put. No path in this PR should be able to lose messages.
- A raw binary log in the ZIP now shadows its stale `.json` twin regardless of ZIP entry order.
- `binaryLogToJson` validates each record's trailing length marker (same as Fix Logs) and stops at the first malformed record, so garbage files convert to nothing instead of nonsense.
- CLI import: per-file error handling, a `filesErrored` count in the JSON report, and exit code 1 when anything failed.
- The export loop only yields to the UI after actually appending a file (previously it could spin on skipped entries).

## To test
- [x] Round trip: export, import into an empty log directory, conversations appear with messages, dates, and correct names (ad-hoc channels and capitalization included)
- [x] Legacy ZIP from 2.2.x/2.3.x: conversations appear; PM names recover their capitalization, channel names recover for recently joined channels
- [x] Heal: re-import over an install with missing `.idx` files (no overwrite), conversations become visible and the summary counts those characters
- [x] Stale `foo.json` artifacts from old CLI imports are gone after re-import, and Fix Logs no longer invents empty conversations from them
- [x] CLI: dry run previews without touching disk; a failed file yields exit code 1
- [x] Backward compat: an old release imports a new export cleanly (sidecar ignored, logs still bare arrays) [Low priority, but just to be safe!]

### DIFF
--- a/electron/services/log-backup.ts
+++ b/electron/services/log-backup.ts
@@ -298,6 +298,7 @@ export function recoverLogName(
   messages: JsonLogMessage[]
 ): string | undefined {
   const key = logKey.split('/').pop()!;
+  if (key === '_') return 'Console';
   if (key.startsWith('#'))
     return context.channelNames.get(character)?.get(key.toLowerCase());
   const match = messages.find(
--- a/electron/services/log-backup.ts
+++ b/electron/services/log-backup.ts
@@ -1,0 +1,380 @@
+/**
+ * @license MPL-2.0
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * @copyright 2024-2026 Sylvia Roselie & Respective Horizon Contributors
+ * @version 1.0
+ * @see {@link https://github.com/Fchat-Horizon/Horizon|GitHub repo}
+ *
+ * Conversion helpers between the binary chat log format, the JSON log format
+ * used inside export ZIPs, and the `.idx` sidecar files the log viewer needs.
+ *
+ * ^ Binary record layout must stay in sync with serializeMessage in
+ *   electron/filesystem.ts:
+ *   u32LE time (seconds) | u8 type | u8 senderLength | sender |
+ *   u16LE textLength | text | u16LE recordLength - 2
+ * ^ Index layout must stay in sync with checkIndex/loadIndex in
+ *   electron/filesystem.ts:
+ *   u8 nameLength | name | 7-byte entries of u16LE day + u40LE offset
+ *
+ * ! NEVER IMPORT ELECTRON HERE! This must ALSO WORK FOR THE CLI!!!!!!!!!!!!!!!!!!!!1
+ */
+
+import fs from 'fs';
+import path from 'path';
+
+/** One message inside a JSON log file. Times are seconds since epoch. */
+export interface JsonLogMessage {
+  time: number;
+  type: number;
+  sender: string;
+  text: string;
+}
+
+/**
+ * One JSON log file inside an export ZIP, normalized.
+ *
+ * ! Exports must stay bare arrays: released clients strip the .json suffix
+ *   and write any other shape verbatim to the binary log path. The optional
+ *   name only tolerates hand-made { name, messages } files on import.
+ */
+export interface JsonLog {
+  name?: string;
+  messages: JsonLogMessage[];
+}
+
+/**
+ * Per-character sidecar mapping log file names to display names. Old
+ * importers ignore it, so new exports stay safe on old clients.
+ */
+export const conversationNamesFile = 'conversation-names.json';
+
+const dayMs = 86400000;
+
+function localDay(timeSeconds: number): number {
+  const date = new Date(timeSeconds * 1000);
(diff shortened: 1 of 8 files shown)


electron/services/log-backup.ts
```diff
@@ -0,0 +1,380 @@
...
+ * recentChannels file, private names from the sender whose name lowercases
+ * to the log's file name.
+ */
+export function recoverLogName(
+  logKey: string,
+  character: string,
+  context: LogImportContext,
+  messages: JsonLogMessage[]
+): string | undefined {
+  const key = logKey.split('/').pop()!;
```
FatCatClient, line 300:
```suggestion
  const key = logKey.split('/').pop()!;
  if (key === "_") return "Console";
```

Unless you also want this to use the localized name.

When importing a 2.3.2 JSON style .zip into an empty directory, the console logs were just labelled "_".

FatCatClient:
Here are the issues I've found after extensive testing:

- Importing a pre-2.3.3 JSON (ie manual) backup breaks the Console tab's name.
- Importing a manual backup generated from [f8ee1c3](https://github.com/Fchat-Horizon/Horizon/pull/893/commits/f8ee1c31fd494b9b75e90761c00d8aabebf06707) doesn't work in 2.3.2.
- I've had a case where the dry run doesn't see any problems, but the _real_ import did:
```
{"op":"import","dryRun":true,"zip":"~/2.3.2_ horizon-export-2026-07-16T19-08-06.zip","touchedCharacters":101,"generalImported":true,"filesErrored":0}
 ~> ./dist/linux-unpacked/horizon-electron import --zip ~/2.3.2\ horizon-export-2026-07-16T19-08-06.zip --data-dir ~/templogs
20:45:09.025 › init.common
20:45:09.028 › init.common.done 
Failed to import characters/(characterName)/logs/#adh-be95b9c642b3612f79c7.json: The value of "value" is out of range. It must be >= 0 and <= 65535. Received 66778
{"op":"import","dryRun":false,"zip":"~/2.3.2 horizon-export-2026-07-16T19-08-06.zip","touchedCharacters":101,"generalImported":false,"filesErrored":1}
```


Just to be on the safe side, I've done additional testing for every case where I tried to work with 2.3.2 manual backups, 2.3.3* manual backups, and auto backups (from 2.3.2). Including importing into empty folders, overwriting a filled folder, and "importing" without overriding into a filled folder.

FatCatClient:
@CodingWithAnxiety 

can you confirm whether all *three* issues in my non-review comment are fixed, or are not important enough to be fixed, not just the first one?

PR #898, opened 2026-07-17
fix: extend the channel prune grace period past the ping threshold

### LINKED ISSUE #897: Channels are pruned from groups too eagerly
### Describe the bug

Although channel pruning (see #848) mostly seems to work fine, there have been some cases reported where this kicks in too early: It can take a while for the F-Chat server to respond with a JCH server command, and by then the prune timeout has already kicked in, resulting in cases where the message shows up to tell you the channel was ungrouped, only for you to actually join the channel anyway.

### Version Information

Phoneposting. It's in 2.3.2 at least, but obviously introduced in 2.3.1.

### To reproduce

1. Have a slow connection to F-Chat and a bunch of grouped channels
2. Connect w/ that characters
3. Roll unlucky on the internet table, and join a channel > 10 seconds after the connection starts
4. Channels are now ungrouped against your will.

### Additional context

My intended solution for this is to simply increase the grace period for channel pruning to be barely above the Ping-Pong client-server check threshold (90 seconds, from the top of my head). That sounds like a lot, but unless there is a freak incident where it takes longer than the `PIN` response for you to get a `JCH` response, the only times you would not get an answer for joined channels within that time period, you would be disconnected from the server anyway.

Then obviously make sure that this timeout is also cancelled when you disconnect for some reason :^)

Kannamoris:
The startup prune from #848 treated any grouped channel that had not joined within a 10 second settle window as deleted, so a slow server response could ungroup channels that were merely late to join. As suggested in the issue, the grace period now sits at the client's ping stale threshold of 90 seconds: a connection that survives that long without a join arriving is demonstrably alive, so the channels that still have not joined are gone for good rather than merely slow. The window still resets on every successful join, and the timer is now also cancelled when the connection closes. Verified against a live connection through a throttling proxy that stalls the server stream for 20 seconds after connect: previously every grouped channel was pruned and then joined anyway, now all channels survive the stall and a fake dead channel is still pruned with the console notice once the grace period runs out.

Fixes #897

### DIFF
--- a/chat/conversations.ts
+++ b/chat/conversations.ts
@@ -38,8 +38,12 @@ const CONVERSATION_CACHE_UPDATE_FREQ_IN_MS = 1000;
  * before pinned channels that never joined are treated as removed and pruned.
  * The timer resets on every successful join, so a slow connection that delivers
  * joins gradually keeps deferring the cleanup until the joins actually stop.
+ * Sits at the client's ping stale threshold (the pin timeout in
+ * fchat/connection.ts): a connection that survives this long without a join
+ * arriving is demonstrably alive, so the channels that still have not joined
+ * are gone for good rather than merely slow.
  */
-const PINNED_CLEANUP_SETTLE_IN_MS = 10000;
+const PINNED_CLEANUP_SETTLE_IN_MS = 90000;
 
 function createMessage(
   this: any,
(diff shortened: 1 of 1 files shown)


CodingWithAnxiety:
Agreed and approved. Merging.

PR #900, opened 2026-07-19
tweak: Add support for species distinctions while maintaining F-List Species Preference matching

saalisu:
While there's some contention about the similarities between the two species, it's generally recognised that naga and lamia are not the same - owing to their different backgrounds and depictions in media. Additionally, it's causing confusion for at least one lamia player being referred to as a 'naga' despite expressly stating otherwise.

I didn't create a lamia grouping because I don't think there's any synonyms or of-like species that fit the description, and I haven't touched the existing Quetzalcoatl entries either because that's a whole third mythos separate from the other two and I don't have much experience in if players of that species would be fine being considered 'naga' for brevity.

This edit was also made directly through GitHub, so there's a non-zero chance there's a formatting change, if that is the case, please let me know, close this PR and I'll do a proper pull in my IDE.

One line edit.


FatCatClient:
> I didn't create a lamia grouping because I don't think there's any synonyms or of-like species that fit the description

Also, since there is no default kink on F-List's end for "lamia" specifically, it would break matching for every character assigned to that group. Or rather, those players couldn't get a score for their species being in someone's preference. 

...Unless you then based the "lamia" group's score on the "Naga" custom kink. Which sounds like the kind of thing you'd not want either.

saalisu:
> ...Unless you then based the "lamia" group's score on the "Naga" custom kink. Which sounds like the kind of thing you'd not want either.

I can't really speak on behalf of people *looking* for characters like that - I would assume they probably would mind either, but as a character *player*, it feels equivocal to lumping skunks and striped polecats together because they *look* similar, but are in fact entirely different families of mammals.

I wasn't fully aware that was the purpose of the grouping, as I'm not actually a Horizon user myself. My issue is more with how the species is displayed. If Lamia and Naga groups shared the same score to be matched with F-List's joint 'Naga/Lamia' species preference kink, that wouldn't be an issue, just as long as the displayed text in the little info element was correct between Lamia and Naga listed profiles.

saalisu:
Okay, I've added `speciesPreferenceMapping` to `matcher-types.ts` and for now added in lamia as a distinct species, yet still bind it to the same `Lamia / Naga` species preference matching ID. I did a quick test and from what I've seen:

1) Lamia and Naga profiles are clearly displayed as distinct species in the li'l profile preview element thingy.
2) The matching still considers 'Lamia' under the 'Lamia / Naga' preference kink, so that still matches up.

Feel free to scrutinise my approach, I hope it isn't too sloppy.

p.s. This should also make it trivial in the future to create more 'visual' distinctions between species while maintaining compatibility with F-Lists rigid list of species. I heard miqo'te players are upset about being called cats or something... :)

FatCatClient:
Code approach looks fine to me, helps that the comments explain it fairly well.

PR #912, opened 2026-07-26
fix: anyAscii improvements

### LINKED ISSUE #497: Any-Ascii can potentially break BBCode formatting in profiles
**Describe the bug**

Some profiles use characters like ``【`` for cool formatting effects like this:

Image

Unfortunately, turning on the "Standardize profile fonts" makes Any-Ascii parse the profile text too, which then means that unicode character gets formatted to a regular square bracket and...

Image

It completely breaks the formatting!

**Ideas**

We have two ways to solve this:

- When a profile's body needs to be parsed through Any-Ascii, we double check for any characters that can be turned into ASCII square brackets and then replace them with something else, so Any-Ascii doesn't return square brackets
- We fork Any-Ascii's code and add the ability to ignore certain characters or character sets with optional parameters. This would also solve the problem of emoji turning into their written out equivalents for text formatting, even though they are likely fine anyway. The Javascript implementation is licensed under the permissible ISC License, so this should be fine.

freenutsxd:
Adds a check when 'Standardize profile fonts' is enabled that will skip any bracket-adjacent or emojis from being converted. This means that bbcode will no longer be broken if used alongside a bracket character like `「」` or `【】` with the setting enabled. All other non-emoji ASCII font should be untouched, but admittedly I don't know any profiles with godawful ASCII on their profile to test with.

### Default (Standardize fonts disabled)
looks fine, because nothing is converted
image

### OLD Standardize fonts enabled
bad, really bad
image

### NEW Standardize fonts enabled
emojis and characters that would be converted to brackets are kept the same. in this example, only the `✓` and `✗` characters were converted
image

Closes #497 .

### DIFF
--- a/site/character_page/character_page.vue
+++ b/site/character_page/character_page.vue
@@ -284,19 +284,32 @@
 
   const standardParser = new StandardBBCodeParser();
 
+  // what anyAscii turns emoji into, e.g. ':emoji:'. nothing below U+2000
+  // converts to this shape, so it can't match real text
+  const ASCII_EMOJI_NAME = /^:[a-z0-9_+-]+:$/;
+
+  // preserves emoji sequences (hands with skin tones, people variations, etc.)
+  const EMOJI_GLUE =
+    /[\u200d\u20e3\ufe0e\ufe0f]|[\u{e0020}-\u{e007f}]|[\u{1f1e6}-\u{1f1ff}]|[\u{1f3fb}-\u{1f3ff}]/u;
+
   // anyAscii turns decorative unicode brackets (「」【】［］) into real ASCII
-  // brackets, which the bbcode parser then tries to convert. this skips those characters
+  // brackets, which the bbcode parser then tries to convert. emojis are also
+  // converted to :emoji: equivalents. this skips those characters
   function safeAnyAscii(s: string): string {
     let out = '';
 
     for (const ch of s) {
-      if (ch.codePointAt(0)! < 128) {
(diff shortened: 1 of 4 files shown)


FatCatClient (dismissed):
Don't forget about kinks and profile infotags, they can have custom user text too-- Even if that doesn't include BBCode.

Slightly off-topic:

Though from my quick testing, I can't get it to parse correctly for kinks consistently-- it seems to really struggle sometimes for certain cached profiles, even if the words themselves should be picked up by Any-Ascii.

But this is likely a regression from either a7eb9390ca8e89879f710da0a5f257da1aa6bc1f or 909fe06a9a523834ab777a8891cc3780d69a0d7a, and obviously not related to your changes.

Back on topic: Your implementation for dealing with characters we don't want to parse with Any-Ascii is good. Probably way smarter than handling it in Any-Ascii itself lmao

freenutsxd:
Wasn't able to reproduce the caching issue in my (limited) testing so I couldn't really investigate it, but I _did_ apply my same changes to infotags and custom kinks.

FatCatClient:
That’s fine, thank you. Those issues are out of scope for this PR anyway.

FatCatClient:
@freenutsxd If you could make those comments in the SafeAscii file proper TSDoc-style comments, so that they work with LSP plugins and later on when we want to export docs to web, that'd be perfect.

Otherwise it's good to merge, I'd say.

freenutsxd:
Done. Lmk if there’s anything I might be missing with the format.

PR #914, opened 2026-07-31
fix: Guestbook counter is accurate past 15 posts

### LINKED ISSUE #854: Guestbook post tab's counter caps at 15
### Describe the bug

Image

This profile has 17 guestbook posts. The first page just shows 15, and that's the amount you see on the profile viewer.

This is really just for the amount listed in the tab. When you check the tab itself and use the paginator, it shows the correct amount.

### Version Information

This bug is so old, I would not be surprised if it's in vanilla 3.0 standalone as well. No need to copy this, it was in Rising too.

### To reproduce

1. Go to profile with over 15 guestbook posts
2. Look at tab

### Additional context

_No response_

freenutsxd:
Fixes an OLD issue where any profiles with more than 15 guestbook posts would only show 15 for the guestbook post counter.  character-guestbook,php never gives us a total guestbook count, so our only option is to walk through every single guestbook page and sum them up.

For most profiles this won't change much, since it's rare to see anyone with more than 15 posts nowadays, but every extra page is an extra API call, so it might be a *little* slower to count some really-filled guestbooks.

Fixes #854

### DIFF
--- a/chat/profile_api.ts
+++ b/chat/profile_api.ts
@@ -504,15 +504,36 @@ async function imagesGet(id: number): Promise<CharacterImage[]> {
   ).images;
 }
 
-async function guestbookGet(id: number, offset: number): Promise<Guestbook> {
-  const data = await core.connection.queryApi<{
+// if anybody has more than 100 *pages* of guestbook posts (1500+ posts), i'll be stunned
+const GUESTBOOK_MAX_COUNT_PAGES = 100;
+
+async function guestbookPageFetch(id: number, page: number) {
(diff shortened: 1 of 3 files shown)


CodingWithAnxiety:
Looks sane. Will merge later.

FatCatClient:
This technically makes the issue where the guestbook post amount is pulled from the cache-- either that, or the numeric indicator does not update through automated refreshes initiated by the profile viewer itself, more obvious. 

But also that's a completely different problem that has no bearing on whether this should be merged or not.

PR #920, opened 2026-08-09
feat: Warn about leading whitespace before sending commands

### LINKED ISSUE #919: Automatically remove space before slash (/) commands
### Describe the problem

In a chat, writing a command like /me, /roll or /bottle with a space (usually, accidentally) entered before the command will send the 'command' as a message instead of the command itself.

### Your idea/ solution

When a message is sent, check for a forward slash command in the first few/several characters of the message; if there are more spaces in the message before the slash, the user will likely notice and remove the spaces anyways. If there is a space, it's removed from the message before the message is sent, leading to the command being sent as a command instead of as a regular chat message. This could be set as an option too, though I imagine most users will leave it enabled by default.

### Additional context

_No response_

freenutsxd:
Removes any leading whitespace from commands before sending them, letting them be actually read as a command rather than sending as raw text. For example, ` /me shrugs` will send the /me command correctly instead of sending the text as-is. Anything that isn't an actual command (` /shrug`) will not be trimmed so that it still sends correctly.

I didn't make a toggle for this to avoid adding even more settings bloat, but I doubt there's many people that intentionally send commands with leading spaces anyways.

Closes #919

### DIFF
--- a/chat/ConversationView.vue
+++ b/chat/ConversationView.vue
@@ -541,7 +541,7 @@
   import ManageChannel from './ManageChannel.vue';
   import MessageView from './message_view';
   import ReportDialog from './ReportDialog.vue';
-  import { isCommand, trimCommandWhitespace } from './slash_commands';
+  import { isCommand } from './slash_commands';
   import UserView from './UserView.vue';
   import CharacterChannelList from './character/CharacterChannelList.vue';
   import * as _ from 'lodash';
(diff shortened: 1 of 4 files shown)


freenutsxd:
Reworked the changes following some feedback. Now whenever a user tries to send a command with a space, there will be a warning at first until the user hits Enter again. I was worried I wasn't going to like this implementation, but honestly, this feels way better to use

image

PR #941, opened 2026-09-06
refactor(about): collect system diagnostics via scoped IPC

CodingWithAnxiety:
Desc. is draft.

This coomit/pr is a "starting base" for working on IPC, apart of a several-stage plan work on IPC. It is intended for 2.4.0/2.4.x

The change for now is small, and scoped. It is a work in progress but this is ready to be merged as-is.

### DIFF
--- a/electron/About.vue
+++ b/electron/About.vue
@@ -5,7 +5,6 @@
     :class="getThemeClass()"
     @auxclick.prevent
   >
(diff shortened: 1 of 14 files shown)


FatCatClient:
Briefly discussed this in private before, but this is a good idea. There are a couple of structural changes that would need to be made if this were ever extended elsewhere.

PR #942, opened 2026-09-07
fix: URL preview allowing popups (for good, probably)

freenutsxd:
Electron's webview has `allowpopup` as a boolean, but no matter what the boolean is set to, it always allows popups if it's included as an attribute at all. Removing `allowpopup` entirely should (hopefully, permanently) prevent any cases of popups. 

Also removed `nodeIntegration` entirely for the same reason as above, which lowkey could've been a security issue.

https://www.electronjs.org/docs/latest/api/webview-tag#allowpopups


CodingWithAnxiety:
Merging due to it being a security PR. Thanks for your contribution!

PR #944, opened 2026-09-08
Native window controls for Windows and Linux

FatCatClient:
Adds native window control overlays for Windows and Linux. MacOS already had these (through different methods), but they look perfectly fine on Linux and Windows too. Doubly so because they add specific functionality (like hover menus on Windows 11) that we can't mimic with our implementation.

~~Consider this a Do not merge for now, because the settings button might float a bit awkwardly. I can align it 1:1 for Windows 11, but on Linux the location where it'd go can vary quite a bit depending on your window decorations. I might remove that button altogether as a result of #640  but I can't say for sure.~~

nvm solved it. because i'm just that good 💅

image

Top: Windows
Bottom: Linux w/ (I think, I can't tell from that install) default KDE window controls.

### DIFF
--- a/electron/Window.vue
+++ b/electron/Window.vue
@@ -139,6 +139,7 @@
           display: flex;
           justify-content: flex-end;
           -webkit-app-region: drag;
+          margin-right: 130px;
         "
         id="windowButtons"
         class="btn-group"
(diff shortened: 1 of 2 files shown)


FatCatClient:
On Windows this has an issue where the margins are set incorrectly when maximising the window. 

I’ll fix that later.

PR #947, opened 2026-09-09
fix: about:blank when opening devtools

### LINKED ISSUE #943: Dev tools no longer work for the Window.vue instance and ImagePreview
### Describe the bug

It used to be so that when you used any of the shortcuts to open dev tools while the app was in non-production mode, it'd open two instances of dev tools: one for the Window context, and one for the web content containing the index root component.

Now it _does_ open both, but the Window's dev tools just show `about:blank`. Clearly they're not attaching properly anymore.

_Edit:_

Verified the same problem also happens with ImagePreview.

### Version Information

I think this started after the Electron 40 upgrade. Don't quote me on that, but it's not a super new issue.

### To reproduce

1. Run horizon in development mode
2. Press Ctrl + Shift + I
3. about:blank'd

### Additional context

_No response_

freenutsxd:
Fixes the dev tools not pointing to the correct window object after some changes that came with Electron 42, causing every instance to show `about:blank` instead. This points the correct view to the dev tools so we actually see something.

I also went and disabled the dev tools button when viewing a preview for any `f-list.net` URL or a character preview, because those have always opened an `about:blank` view due to how we handle those URLs. @FatCatClient idk if you want to/already did handle that differently in the V2 previewer, but I figured you should know.

This should *probably* be merged for 2.4.0 since it's a regression 

Closes #943

### DIFF
--- a/chat/preview/ImagePreview.vue
+++ b/chat/preview/ImagePreview.vue
@@ -7,7 +7,7 @@
     <div class="image-preview-toolbar" v-show="sticky || debug">
       <a
         @click="toggleDevMode()"
-        :class="{ toggled: debug }"
+        :class="{ toggled: debug, disabled: !debug && !canDebug }"
         :title="l('imagePreview.debug')"
         ><i class="fa fa-terminal"></i
       ></a>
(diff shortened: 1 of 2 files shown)


FatCatClient:
> I also went and disabled the dev tools button when viewing a preview for any f-list.net URL or a character preview, because those have always opened an about:blank view due to how we handle those URLs. @FatCatClient idk if you want to/already did handle that differently in the V2 previewer, but I figured you should know.

I didn't really take that into consideration at all, so it's fine if we merge that with this PR anyway.

